PR #1, opened 2023-06-13
feat:Database configuration

KubaBaniak:


src/app.module.ts
```diff
@@ -1,9 +1,28 @@
...
+        port: +configService.get<number>('DB_PORT'),
+        username: configService.get('DB_USERNAME'),
+        password: configService.get('DB_PASSWORD'),
+        database: configService.get('DB_NAME'),
+        entities: [],
+        synchronize: false,
+      }),
+      inject: [ConfigService],
+    }),
+  ],
```
rad20c, line 25: Let's go with Prisma instead of TypeOrm 🙂 -> check the docs here: https://docs.nestjs.com/recipes/prisma

docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+    environment:
+      PGADMIN_DEFAULT_EMAIL: admin@admin.com
+      PGADMIN_DEFAULT_PASSWORD: root
+    ports:
+      - "5050:80"
+    profiles:
+      - tools
+volumes:
+  db:
+    driver: local
```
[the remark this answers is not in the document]
rad20c: But overall good job at docker-compose 👍🏻

todo.txt
```diff
@@ -1,3 +1,6 @@
-baza danych postgress
```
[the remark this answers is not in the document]
rad20c: hmm, even though you have added `todo.txt` to the `.gitignore` file, this file is still present - double check if you have commited removal of this file from the repo. Perhaps to fix this you will have to:
- temporarily remove `todo.txt` entry from `.gitignore`
- backup `todo.txt` somewhere on your disk
- remove `todo.txt` file
- commit change (remove `todo.txt` file from repository)
- bring back `todo.txt` to `.gitignore`

The above should remove file from the repository - however they won't remove `todo.txt` from the history - for this you can try squash + force merge approach

docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+    ports:
+      - '5432:5432'
+    volumes: 
+      - db:/var/lib/postgresql/data
+  pgadmin:
+    image: dpage/pgadmin4
+    restart: always
+    environment:
+      PGADMIN_DEFAULT_EMAIL: admin@admin.com
+      PGADMIN_DEFAULT_PASSWORD: root
```
rad20c, line 18: consider reading these from the `.env` file - we don't have to do this right now in this PR, but sooner or later we should change it

rad20c: also once this is done we'll have to use different credentials as these are already committed and should be treated as leaked

rad20c (approved):
Looks good to me 👍🏻 Left some additional comments but I'm giving an approval already, so once you read the comments and apply (or not) the suggestions you can merge this PR 👍🏻

PR #2, opened 2023-06-14
Feat/User's CRUD and authentication

KubaBaniak:


src/auth/auth.controller.ts
```diff
@@ -0,0 +1,19 @@
+import { Controller, Body, Post, HttpCode, HttpStatus } from '@nestjs/common';
+import { AuthService } from './auth.service';
+
+@Controller('auth')
+export class AuthController {
+  constructor(private authService: AuthService) {}
+
+  @HttpCode(HttpStatus.OK)
+  @Post('login')
+  async signIn(@Body() userData: Record<string, any>) {
```
rad20c, line 10: in routes like this try to keep consistency between route / controller naming. Here we have a `login` word in route. In this case the controller method should be `login`

rad20c: you can also change the route to `signin`

src/auth/auth.controller.ts
```diff
@@ -0,0 +1,19 @@
...
+    return this.authService.signIn(userData.email, userData.password);
+  }
+
+  @HttpCode(HttpStatus.OK)
+  @Post('signup')
+  async singupUser(@Body() userData: Record<string, any>) {
```
rad20c, line 16: same as above - use specific DTO interface / class (with class-validator you most probably would want to use class, if not using class-validator interface should be fine)

src/auth/auth.service.ts
```diff
@@ -0,0 +1,39 @@
...
+import { User } from '@prisma/client';
+
+@Injectable()
+export class AuthService {
+  constructor(
+    private userService: UserService,
+    private jwtService: JwtService,
+  ) {}
+
+  async signIn(email: string, pass: string): Promise<any> {
```
rad20c, line 17: from what I see we are returning `Promise<{ access_token: string }>` here. Replace `any` with a proper return type. Also you can declare the type (or interface) either in a separate file or somewhere in this service. I'd vote for a separate file

.gitignore
```diff
@@ -8,6 +8,9 @@
 #notes
 todo.txt
 
+#db
+/prisma/migrations
```
rad20c, line 12: I think this shouldn't be ignored, it should be commited as it contains a history of migrations (changes in db schema). Please confirm with Rafał Ch.

src/app.module.ts
```diff
@@ -2,10 +2,17 @@ import { Module } from '@nestjs/common';
...
+import { AuthService } from './auth/auth.service';
+import { LocalAuthGuard } from './auth/guards/local-auth.guard';
 
 @Module({
-  imports: [PrismaModule],
-  controllers: [AppController],
-  providers: [AppService],
+  imports: [PrismaModule, AuthModule, UserModule],
+  controllers: [AppController, UserController, AuthController],
+  providers: [AppService, UserService, AuthService, LocalAuthGuard],
```
rad20c, line 16: You should not need to add `UserController` or `AuthController` - `AppModule` is not using it. Same for `UserService` and `AuthService`. And I think `LocalAuthGuard` too but I'm not sure. They should be added to the feature modules they are used in, e.g. `UserController` should be added to the `UserModule` only. `AuthService` in `AuthModule` etc.

prisma/schema.prisma
```diff
@@ -13,18 +13,8 @@ datasource db {
 model User {
   id         Int         @id @default(autoincrement())
   email      String      @unique
-  password   String
+  password   String
```
rad20c, line 16: autoformatter added this?

KubaBaniak: I guess so

src/auth/auth.module.ts
```diff
@@ -0,0 +1,26 @@
...
+import { LocalAuthGuard } from './guards/local-auth.guard';
+
+@Module({
+  imports: [
+    UserModule,
+    PassportModule,
+    JwtModule.register({
+      global: true,
+      secret: jwtConstants.secret,
+      signOptions: { expiresIn: '30000s' },
```
rad20c, line 19: Let's say for now it's ok, but later we should generate a pair of tokens: access token and a refresh token. Then access token will be a short lived token - e.g. 5-30 minutes and refresh token will have a longer expiry date - e.g. 24h / 7 days / 30 days

src/auth/auth.controller.ts
```diff
@@ -0,0 +1,24 @@
...
+import { AuthGuard } from '@nestjs/passport';
+
+@Controller('auth')
+export class AuthController {
+  constructor(private authService: AuthService) {}
+
+  @HttpCode(200)
+  @UseGuards(LocalAuthGuard)
+  @Post('signin')
+  async signIn(@Body() user: User) {
```
rad20c, line 15: include a response type too. Here `signIn` method is `async` and `this.authService.signIn` returns `Promise<{ access_token: string }>` -> the return type should be `Promise<{ access_token: string }>` so:

```suggestion
  async signIn(@Body() user: User): Promise<{ access_token: string }> {
```
but, I think I already suggested somewhere int this PR to move this type to a separate file so that we can do:
```suggestion
  async signIn(@Body() user: User): Promise<SomeExampleSignInResponseType> {
```

rad20c: Looking further into this, from `signIn` method you can return `String` for now instead of `{ access_token: string }` and here in controller map the response to a DTO response type. You should create a `SignInResponseDto` class like this:

```typescript
export class SignInResponseDto {
  constructor(public accessToken: string) {}

  public static from(accessToken: string): SignInResponseDto {
    return new SignInResponseDto(accessToken);
  }
}
```

and here in controller do:
```
const authServiceResult = await this.authService.signIn(user);

return SignInResponseDto.from(authServiceResult);
```

src/auth/auth.service.ts
```diff
@@ -0,0 +1,47 @@
...
+    private userService: UserService,
+    private jwtService: JwtService,
+  ) {}
+
+  async signIn(user: User): Promise<{ access_token: string }> {
+    const payload = { email: user.email, password: user.password };
+    return {
+      access_token: await this.jwtService.signAsync(payload),
+    };
+  }
```
rad20c, line 25: If you follow my comment above ☝🏻 , then you can change it to:
```suggestion
 signIn(user: User): Promise<string> {
    const payload = { email: user.email, password: user.password };
    return this.jwtService.signAsync(payload);
  }
```

Note the lack of `await` next to `this.jwtService.signAsync(payload)` - that's because we can await this one level higher - in controller.  I also removed `async` as it's now not needed (no `await`'s). Note that the return type is still `Promise`, because even without `await` we will get `Promise` here, just a pending one (that will be resolved one level higher).

package.json
```diff
@@ -36,9 +45,12 @@
     "@nestjs/cli": "9.0.0",
     "@nestjs/schematics": "9.0.0",
     "@nestjs/testing": "9.0.0",
+    "@types/bcrypt": "5.0.0",
     "@types/express": "4.17.13",
     "@types/jest": "29.5.1",
     "@types/node": "18.16.12",
+    "@types/passport-jwt": "3.0.8",
+    "@types/passport-local": "1.0.35",
```
rad20c, line 53: good job at differentiating between dependencies and devDependencies 👍🏻

src/auth/auth.controller.ts
```diff
@@ -0,0 +1,24 @@
...
+    return await this.authService.signIn(user);
+  }
+
+  @HttpCode(201)
+  @Post('signup')
+  async singUp(@Body() dto: AuthDto) {
```
rad20c, line 21: for this method also check the above comment

rad20c: also one more thing - I would use separate request dtos for `signIn` and `signUp` method. On a first glance were introducing redundancy but on the other hand we are gaining decoupling. So here we would have e.g.
```suggestion
  async singUp(@Body() signUpRequest: SignUpRequestDto) {
```

rad20c: also, noticed just now - there is a typo in `singUp` - should be `signUp` :)

src/auth/dto/index.ts
```diff
@@ -0,0 +1 @@
+export * from './auth.dto';
```
rad20c, line 1: good idea on using `index.ts` with exports 👍🏻

src/user/user.controller.ts
```diff
@@ -0,0 +1,14 @@
...
+  Controller,
+  Get,
+  Param,
+  Post,
+  Body,
+  Put,
+  Delete,
+} from '@nestjs/common';
+import { UserService } from './user.service';
+import { User as UserModel } from '@prisma/client';
```
rad20c, line 11: Hmm, I think controller should not have imports to prisma - see if you will be able to remove this import after you have implemented the approach I've mentioned above about response DTOs classes

src/user/user.service.ts
```diff
@@ -0,0 +1,68 @@
+import { Injectable } from '@nestjs/common';
+import { PrismaService } from 'src/prisma/prisma.service';
+import { User, Prisma } from '@prisma/client';
+
+@Injectable()
+export class UserService {
+  constructor(private prisma: PrismaService) {}
+
+  async findOneUser(email: string): Promise<User | undefined> {
```
rad20c, line 9: normally I'd prefer a service to throw a "Not Found` error but let's skip it as this is more a dao than a service

src/auth/strategies/jwt.strategy.ts
```diff
@@ -0,0 +1,20 @@
...
+    super({
+      jwtFromRequest: ExtractJwt.fromAuthHeaderAsBearerToken(),
+      ignoreExpiration: false,
+      secretOrKey: jwtConstants.secret,
+    });
+  }
+
+  async validate(dto: AuthDto) {
+    return dto;
+  }
```
rad20c, line 19: hmm, is this unfinished? I think you should check the `jwt` token here if it's a valid one (it's not expired and signature matches)

src/auth/strategies/local.strategy.ts
```diff
@@ -0,0 +1,22 @@
...
+    super({ usernameField: 'email' });
+  }
+
+  async validate(email: string, password: string): Promise<User> {
+    const data = { email, password };
+    const user = await this.authService.validateUser(data);
+    if (!user) {
+      throw new UnauthorizedException();
+    }
+    return user;
```
rad20c, line 20: I like to add new space before code that opens `{}` like e.g. `if () {...}` and before `return`, like this:
```suggestion
    const user = await this.authService.validateUser(data);
    
    if (!user) {
      throw new UnauthorizedException();
    }
    
    return user;
```

but I know developers have their own preferences so I'm not forcing you to change it 🙂

rad20c (changes requested):
Overall good job 👍🏻 See my comments in the review 👍🏻

src/auth/strategies/jwt.strategy.ts
KubaBaniak: Don't `ignoreExpiration: false` and `secretOrKey: jwtConstants.secret` check if it's a valid one?

rad20c: Hmm, I think these are just configuration options. Let's do it this way: 

- currently you are not using this strategy anywhere so let's leave it as it is for now
- in the next tasks that you have ahead you will need to use this strategy
- once you implement endpoint using this strategy test / verify how it works (e.g. modify manually JWT token signature so it's invalid or alter expiration date)
- I think that using this strategy now the invalid token will pass, but I might be wrong

src/auth/auth.controller.ts
```diff
@@ -0,0 +1,24 @@
+import { Controller, Body, Post, HttpCode, UseGuards } from '@nestjs/common';
+import { AuthService } from './auth.service';
+import { AuthDto } from './dto';
+import { LocalAuthGuard } from './guards/local-auth.guard';
+import { User } from '@prisma/client';
+import { AuthGuard } from '@nestjs/passport';
+
+@Controller('auth')
+export class AuthController {
+  constructor(private authService: AuthService) {}
```
rad20c, line 9: this can be readonly:

```suggestion
  constructor(private readonly authService: AuthService) {}
```

src/auth/auth.controller.ts
```diff
@@ -0,0 +1,35 @@
+import { Controller, Body, Post, HttpCode, UseGuards } from '@nestjs/common';
+import { AuthService } from './auth.service';
+import { LocalAuthGuard } from './guards/local-auth.guard';
+import {
+  SignInRequestDto,
+  SignInResponseDto,
+  SignUpRequestDto,
+  SignUpResponseDto,
```
rad20c, line 8: let's experiment a little and drop the `Dto` - just name it `SignInRequest`, `SignInResponse` etc.

src/auth/strategies/jwt.strategy.ts
```diff
@@ -0,0 +1,20 @@
...
+export class JwtStrategy extends PassportStrategy(Strategy) {
+  constructor() {
+    super({
+      jwtFromRequest: ExtractJwt.fromAuthHeaderAsBearerToken(),
+      ignoreExpiration: false,
+      secretOrKey: jwtConstants.secret,
+    });
+  }
+
+  async validate(payload: any) {
```
rad20c, line 17: add return type

rad20c: as discussed - we will do it in the next PR

src/auth/dto/validate-user-request.ts
```diff
@@ -0,0 +1,12 @@
+import { IsEmail, IsNotEmpty, IsString, MinLength } from 'class-validator';
+
+export class ValidateUserRequest {
```
rad20c, line 3:
```suggestion
export class UserRequest {
```

rad20c (approved):
LGTM

PR #6, opened 2023-06-19
feat: update recipe

KubaBaniak:
[Task description](https://app.clickup.com/t/862jzjdhv])

### DIFF
--- a/src/recipe/dto/update-recipe-response.ts
+++ b/src/recipe/dto/update-recipe-response.ts
@@ -1,0 +1,14 @@
+interface IUpdatedRecipe {
+  title: string;
+  description: string;
+  ingredients: string;
+  preparation: string;
+}
+
+export class UpdatedRecipeResponse {
+  constructor(public updatedRecipes: IUpdatedRecipe) {}
+
+  public static from(updatedRecipes: IUpdatedRecipe): UpdatedRecipeResponse {
+    return new UpdatedRecipeResponse(updatedRecipes);
+  }
+}
--- a/src/recipe/dto/update-recipe-response.ts
+++ b/src/recipe/dto/update-recipe-response.ts
@@ -1,14 +1,23 @@
-interface IUpdatedRecipe {
-  title: string;
-  description: string;
-  ingredients: string;
-  preparation: string;
-}
+import { Recipe } from '@prisma/client';
 
 export class UpdatedRecipeResponse {
-  constructor(public updatedRecipes: IUpdatedRecipe) {}
+  constructor(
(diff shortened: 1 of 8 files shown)


src/recipe/dto/update-recipe-response.ts
```diff
@@ -0,0 +1,14 @@
...
+  title: string;
+  description: string;
+  ingredients: string;
+  preparation: string;
+}
+
+export class UpdatedRecipeResponse {
+  constructor(public updatedRecipes: IUpdatedRecipe) {}
+
+  public static from(updatedRecipes: IUpdatedRecipe): UpdatedRecipeResponse {
```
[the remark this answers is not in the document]
rad20c: with this approach I think we will get a response like this:

```json
{
  updatedRecipes: {
    title: string;
    description: string;
    ingredients: string;
    preparation: string;
  }
}
```

I think this additional layer of nesting in this case is not needed, we should change it so that we get a result like this:

```json
{
  title: string;
  description: string;
  ingredients: string;
  preparation: string;
}
```

src/recipe/recipe.controller.ts
```diff
@@ -41,11 +44,24 @@ export class RecipeController {
...
   @Get()
+  @UseGuards(JwtAuthGuard)
   async fetchRecipes(): Promise<FetchRecipesResponse> {
     const fetchedRecipes = await this.recipeService.fetchAllRecipes();
 
     return FetchRecipesResponse.from(fetchedRecipes);
   }
+
+  @UseGuards(JwtAuthGuard)
+  @Patch(':id')
```
rad20c, line 58: quick question: how does the update work right now? Is it updating (replacing) the whole object or is it updating only the fields we provide in the request?

- If the first approach then this should be a `@Put`
- If the second approach then `@Patch` is fine 👍🏻 (I believe it's this one in this case)

src/recipe/recipe.service.ts
```diff
@@ -24,8 +25,24 @@ export class RecipeService {
...
+  ): Promise<Recipe> {
+    try {
+      const recipe = await this.prisma.recipe.update({
+        where: { id },
+        data: payload,
+      });
+      return recipe;
+    } catch {
+      throw new NotFoundException();
+    }
```
rad20c, line 46: yep, in this case awaiting is justified by the try catch block 👍🏻 good job 👍🏻 

Later when we introduce repository / dao we will probably return prisma call immediately and then in service await and try catch it

rad20c (approved):

PR #7, opened 2023-06-20
feat: delete recipe

KubaBaniak:
### [Task description](https://app.clickup.com/t/862jzje1n)

### DIFF
--- a/src/recipe/recipe.controller.ts
+++ b/src/recipe/recipe.controller.ts
@@ -8,6 +8,7 @@ import {
   Get,
   ParseIntPipe,
   Patch,
+  Delete,
 } from '@nestjs/common';
 import { RecipeService } from './recipe.service';
 import {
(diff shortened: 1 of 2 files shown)


src/recipe/recipe.controller.ts
```diff
@@ -64,4 +65,10 @@ export class RecipeController {
 
     return UpdatedRecipeResponse.from(updatedRecipe);
   }
+
+  @UseGuards(JwtAuthGuard)
+  @Delete(':id')
+  async deleteRecipe(@Param('id', ParseIntPipe) id: number) {
```
PrzerostNadSens, line 71: please add what type this function returns

src/recipe/recipe.service.ts
```diff
@@ -45,4 +49,14 @@ export class RecipeService {
       throw new NotFoundException();
     }
   }
+
+  async deleteRecipe(id: number) {
```
PrzerostNadSens, line 53: please add what type this function returns

src/recipe/recipe.service.ts
```diff
@@ -45,4 +49,14 @@ export class RecipeService {
...
+    try {
+      await this.prisma.recipe.delete({
```
PrzerostNadSens, line 55: it is ok but you should think about softDelete because we prefer to mark data in DB as deleted than really delete something

PrzerostNadSens: to implement softDelete you should add an optional parameter to your `Recipe` model `deletedAt` and type `DateTime`

and then you have to take care to not returned softDeleted recipes in get recipe endpoint

PrzerostNadSens: but implementing `softDelete` is optional now and you can skip it

PrzerostNadSens (approved):

PR #9, opened 2023-06-21
feat: api documentation

KubaBaniak:
## [Add API endpoints documentation](https://app.clickup.com/t/862k0a980)


nest-cli.json
```diff
@@ -1,8 +1,19 @@
...
+    "sourceRoot": "src",
+    "compilerOptions": {
+        "deleteOutDir": true,
+        "plugins": [
+            {
+                "name": "@nestjs/swagger",
+                "options": {
+                    "dtoFileNameSuffix": [
+                        "-request.ts",
+                        "-response.ts"
```
rad20c, line 13: hmm, I didn't take this into account when dropping the `.dto` 🤔 Let's leave it for now, we will see later if it brings us some trouble or not

src/auth/dto/sign-up-response.ts
```diff
@@ -1,9 +1,19 @@
...
+  public email: string;
+  public password: string;
+  public role: string;
+
+  constructor(id: number, email: string, password: string, role: string) {
+    this.id = id;
+    this.email = email;
+    this.password = password;
+    this.role = role;
+  }
```
rad20c, line 14: I guess you are doing it this way so that the DTOs are properly displayed in the Swagger? Have you searched for a solution to have Swagger read DTOs correctly and define the fields only in `constructor`?

src/main.ts
```diff
@@ -1,14 +1,26 @@
...
+
   app.useGlobalPipes(
     new ValidationPipe({
       whitelist: true,
     }),
   );
+
+  const config = new DocumentBuilder()
+    .setTitle('Recipe App')
+    .setDescription('The recipe app API description')
```
rad20c, line 17: please set a better description 😅

src/main.ts
```diff
@@ -1,14 +1,26 @@
...
+    .setVersion('1.0')
+    .addTag('Recipes')
```
rad20c, line 19: not sure if you need this tag - this might be responsible for showing this, which does not expose anything:
image

src/main.ts
```diff
@@ -1,14 +1,26 @@
...
+    .build();
+  const document = SwaggerModule.createDocument(app, config);
+  SwaggerModule.setup('api', app, document);
+
```
rad20c, line 23: consider moving the swagger configuration to a separate config file, import this file here and call some method like `initializeSwagger()` or `configureSwagger()` or something similar

src/recipe/dto/fetch-recipes-response.ts
```diff
@@ -1,16 +1,19 @@
...
 
 export class FetchRecipesResponse {
-  constructor(public fetchedRecipes: IFetchedRecipes[]) {}
+  public fetchedRecipes: FetchedRecipe[];
+  constructor(fetchedRecipes: FetchedRecipe[]) {
+    this.fetchedRecipes = fetchedRecipes;
+  }
 
-  public static from(fetchedRecipes: IFetchedRecipes[]): FetchRecipesResponse {
+  public static from(fetchedRecipes: FetchedRecipe[]): FetchRecipesResponse {
```
rad20c, line 16: I think I already commented on this: don't use `FetchedRecipe` as input parameter - use the one that you are mapping from, this will probably be some prisma class

src/recipe/recipe.controller.ts
```diff
@@ -20,8 +20,10 @@ import {
   UpdateRecipeRequest,
 } from './dto';
 import { JwtAuthGuard } from 'src/auth/guards/jwt-auth.guard';
+import { ApiTags } from '@nestjs/swagger';
 
 @Controller('recipe')
+@ApiTags('recipe')
```
rad20c, line 26: - route should be in plural form - `recipes`
- API tag should be e.g. `Recipes`

src/auth/dto/sign-up-response.ts
KubaBaniak: Yes I've searched for solution to this problem, but I couldn't find anything what can solve that.

src/recipe/recipe.controller.ts
rad20c: I think you can leave it as `Auth` 👍🏻 These tags are cosmetics really - make it so that it looks good in the swagger docs page. Just thought that if it starts with a capital letter it will look better :)

src/auth/dto/sign-up-response.ts
rad20c: ok, I guess we will leave it like that then. I guess we would probably have to define it like this either way if we want to use `@ ApiProperty` (not sure about this though)

src/main.ts
```diff
@@ -1,14 +1,22 @@
...
 async function bootstrap() {
   const app = await NestFactory.create(AppModule);
+
   app.useGlobalPipes(
     new ValidationPipe({
       whitelist: true,
     }),
   );
+
+  const swagger_config = initalizeSwagger();
```
rad20c, line 16: use `camelCase`

rad20c, line 16:
```suggestion
  const swaggerConfig = initalizeSwagger();
```

package.json
```diff
@@ -42,6 +43,7 @@
     "typeorm": "0.3.16"
   },
   "devDependencies": {
+    "@faker-js/faker": "8.0.2",
```
rad20c, line 46: is faker used in this PR?

rad20c (approved):

PR #10, opened 2023-06-21
refactor: explicitly define fields in request / response DTO classes

KubaBaniak:
## [Explicitly define fields in request / response DTO classes](https://app.clickup.com/t/862k0ab5m)


rad20c (approved):

rad20c:
One thing - instead of `Task description` you could paste here a task title - in this case `Explicitly define fields in request / response DTO classes`

PR #11, opened 2023-06-22
refactor: update bcrypt config

KubaBaniak:
## [Update Bcrypt configuration to use salt rounds](https://app.clickup.com/t/862k0aa1v)


rad20c (approved):
approved ✅ 

FYI instead of `Number(value)` you can also do `+value` but `Number(value)` is good 👍🏻

PR #14, opened 2023-06-26
test: unit tests

KubaBaniak:
## [Write unit tests](https://app.clickup.com/t/862k0afah)


package.json
```diff
@@ -1,91 +1,98 @@
...
+    "description": "",
+    "author": "",
+    "private": true,
+    "license": "UNLICENSED",
+    "scripts": {
+        "build": "nest build",
+        "format": "prettier --write \"src/**/*.ts\" \"test/**/*.ts\"",
+        "start": "nest start",
+        "start:dev": "nest start --watch",
+        "start:debug": "nest start --debug --watch",
```
rad20c, line 13: do you know why your indentation changed from 2 to 4 lines? If it was intended - avoid doing this in PR related to some other task. If not - you should probably check why it happened.

KubaBaniak: The only thing that I changed here was Jest configuration in here. There was a problem with importing modules and this fixed this issue. https://github.com/KubaBaniak/RecipeAppNestjs/blob/8ad0783889af8476e81b004b0151c1d321b8e15d/package.json#L71-L84

src/recipe/recipe.controller.spec.ts
```diff
@@ -1,18 +1,175 @@
...
+        });
+    });
+    describe('DeleteRecipe', () => {
+        it('should delete recipe with given id', async () => {
+            //given
+            const id = '1';
+
+            //when
+            const deletedRecipe = await recipeController.deleteRecipe(+id);
+            console.log(deletedRecipe);
```
rad20c, line 169: remove console.log

src/recipe/recipe.controller.spec.ts
```diff
@@ -1,18 +1,175 @@
...
+            //then
+            expect(createdRecipe).toEqual({
+                ...mockRecipe,
+            });
+        });
+    });
+    describe('FetchRecipe', () => {
+        it('should fetch recipe given id', async () => {
+            //given
+            const id = '1';
```
rad20c, line 122: why `string`? `fetchRecipe` expects `id` as `number` - just pass a number

src/recipe/recipe.controller.spec.ts
```diff
@@ -1,18 +1,175 @@
...
+
+            //when
+            const fetchedRecipe = await recipeController.fetchRecipe(+id);
```
rad20c, line 125: if you pass a number then you don't have to parse it here:
```suggestion
            const fetchedRecipe = await recipeController.fetchRecipe(id);
```

src/recipe/recipe.controller.spec.ts
```diff
@@ -1,18 +1,175 @@
...
+            expect(fetchedAllRecipes.fetchedRecipes).toEqual(
+                expect.arrayContaining([mockRecipe]),
+            );
+        });
+    });
+    describe('UpdateRecipe', () => {
+        it('should update recipe with given credentials', async () => {
+            //given
+            const id = '1';
+            const dto = {
```
rad20c, line 148: hmm, perhaps a `request` name would be better?
```suggestion
            const request = {
```

src/recipe/recipe.controller.spec.ts
```diff
@@ -1,18 +1,175 @@
...
-  });
+    const mockRecipeService = {
+        createRecipe: jest.fn((dto: mockRecipeRequest) => {
+            return {
+                id: 1,
+                createdAt: Date.now(),
+                ...dto,
+            };
+        }),
+        fetchRecipe: jest.fn((id: number) => {
```
rad20c, line 35: hmm, if you want to mock the entire recipe service this way, then I think it's better to create a separate file for it and put it in a class, e.g. `MockRecipeService` or `RecipeServiceMock`.

src/recipe/recipe.controller.spec.ts
```diff
@@ -1,18 +1,175 @@
...
+            return {
+                id,
+                createdAt: Date.now(),
+                title: dto.title || 'latest title',
+                description: dto.description || 'latest description',
+                ingredients: dto.ingredients || 'latest ingredients',
+                preparation: dto.preparation || 'latest preparation',
+            };
+        }),
+        deleteRecipe: jest.fn((_id: number) => {
```
rad20c, line 75: why `_id` instead of `id`?

src/recipe/recipe.controller.spec.ts
KubaBaniak: I though `id` as `string` better mimics, what is happening in controller (it prases `string` to `int` via `ParseIntPipe`) and i thought it's some sort of necessary

src/recipe/recipe.controller.spec.ts
```diff
@@ -1,18 +1,175 @@
...
+            return Promise<void>
```
rad20c, line 76: hmm this look kind of strange - `Promise<void>` is a type, not something we return. If it works then ok, but perhaps this could be better:
```typescript
deleteRecipe: jest.fn()
```

or:
```typescript
deleteRecipe: jest.fn((id: number) => {
  return Promise.resolve();
})
```

or:
```typescript
deleteRecipe: jest.fn((id: number) => Promise.resolve())
```

src/recipe/recipe.controller.spec.ts
```diff
@@ -1,18 +1,175 @@
...
+        updateRecipe: jest.fn((id: number, dto: mockRecipeRequest) => {
+            return {
+                id,
+                createdAt: Date.now(),
+                title: dto.title || 'latest title',
+                description: dto.description || 'latest description',
+                ingredients: dto.ingredients || 'latest ingredients',
+                preparation: dto.preparation || 'latest preparation',
+            };
+        }),
```
rad20c, line 74: as the original method is `async` - double check if you shouldn't return a `Promise` here:
```suggestion
        updateRecipe: jest.fn((id: number, dto: mockRecipeRequest) => {
            return Promise.resolve({
                id,
                createdAt: Date.now(),
                title: dto.title || 'latest title',
                description: dto.description || 'latest description',
                ingredients: dto.ingredients || 'latest ingredients',
                preparation: dto.preparation || 'latest preparation',
            });
        }),
```

you can test it by trying to call the method within the test without `await` - if it returns correctly mocked data instead of a pending response then perhaps we should do the above

src/recipe/recipe.controller.spec.ts
KubaBaniak: I wanted to mark it as 'unused value' but i couldn't find any good convention on internet for this. So i sticked with Rust one tbh

src/recipe/recipe.controller.spec.ts
```diff
@@ -1,18 +1,175 @@
...
-    expect(controller).toBeDefined();
-  });
+    const mockRecipeService = {
+        createRecipe: jest.fn((dto: mockRecipeRequest) => {
+            return {
+                id: 1,
+                createdAt: Date.now(),
+                ...dto,
+            };
+        }),
```
rad20c, line 34: you can shorten this method (but don't have to!) like this:
```suggestion
        createRecipe: jest.fn((dto: mockRecipeRequest) => ({
                id: 1,
                createdAt: Date.now(),
                ...dto,
            })
        ),
```

rad20c (commented):
So far it looks good! add service tests now - they are in fact more important than controller tests as they include business logic!

rad20c:
Also, once you are done with recipe tests, don't forget about auth and user tests!

src/recipe/recipe.controller.spec.ts
rad20c: I see, but I think that we can provide `number` right away as `ParseIntPipe` is not triggered in this test (otherwise we would be able to pass a `string` to the controller method)

src/recipe/recipe.controller.spec.ts
rad20c: ah I see now - yes adding underscore at the beginning is in fact they correct way to mark the variable as unused in the TypeScript 👍🏻

ok, then let's do it this way: check out the comment below - if you decide to go with `deleteRecipe: jest.fn()` then problem solved 👍🏻 If you decide with other option (with implementation) go ahead and use `_id` 👍🏻

src/recipe/recipe.controller.spec.ts
rad20c: In the original method we return a Promise, so I'd advise to return a Promise in a mocked method too to be consistent. And yes, all other methods would have to be updated.

It would seem like with current approach we can mock async method with a non-async one, but wuth using jest.spyOn I guess this won't work

package.json
```diff
@@ -1,91 +1,99 @@
...
-      "**/*.(t|j)s"
-    ],
-    "coverageDirectory": "../coverage",
-    "testEnvironment": "node"
-  },
-  "prisma": {
-    "schema": "prisma/schema.prisma"
-  }
+    "dependencies": {
+        "@faker-js/faker": "8.0.2",
```
rad20c, line 24: you can move `@faker-js/faker` to `devDependencies - you don't need this on production

rad20c: hint: you can install libraries to `devDependencies` by using `--save-dev` or `-D`, e.g. `npm i @faker-js/faker -D`

src/recipe/recipe.controller.spec.ts
```diff
@@ -1,18 +1,169 @@
 import { Test, TestingModule } from '@nestjs/testing';
+import { RecipeService } from './recipe.service';
 import { RecipeController } from './recipe.controller';
+import { PrismaService } from 'src/prisma/prisma.service';
+import { faker } from '@faker-js/faker';
 
-describe('RecipeController', () => {
-  let controller: RecipeController;
+describe("Recipe Controller's tests", () => {
```
rad20c, line 7: I'd leave `RecipeController`:

when you run the tests you will have something like this:

```
RecipeController
  should do this
  should do that
```

we're testing RecipeController here, not RecipeController's tests 😄

src/user/user.service.spec.ts
```diff
@@ -1,18 +1,127 @@
...
 describe('UserService', () => {
-  let service: UserService;
+  let userService: UserService;
+
+  const mockUser = {
+    id: expect.any(Number),
+    email: expect.any(String),
+    password: expect.any(String),
+    role: expect.any(String),
+  };
```
rad20c, line 14: hmm, this does not look like a mock for me - more like a payload for an `expect` function

mock should have some values, e.g.:

```suggestion
  const mockUser = {
    id: 1234,
    email: faker.internet.email(),
    password: faker.internet.password(),
    role: 'USER'
  };
```

what you have here are instructions what values to expect in `expect` chain

src/recipe/recipe.service.spec.ts
```diff
@@ -1,18 +1,188 @@
...
+
+      //then
+      expect(updatedRecipe).toEqual(mockRecipe);
+    });
+  });
+
+  describe('DeleteRecipe', () => {
+    it('should delete recipe by given id', async () => {
+      //given
+      const id = 37;
```
rad20c, line 179: if id does not play important role here, I'd set a random id each time (you can use faker or Math.random for example)

src/recipe/recipe.service.spec.ts
```diff
@@ -1,18 +1,188 @@
...
+
+      //then
+      expect(fetchedRecipe).toEqual({
+        id: expect.any(Number),
+        createdAt: expect.any(Number),
+        title: expect.any(String),
+        description: expect.any(String),
+        ingredients: expect.any(String),
+        preparation: expect.any(String),
+      });
```
rad20c, line 141: preferably use `expect.any` if you really don't know what the value of some field will be. But if you know, compare with that value (e.g. if you know a certain mock will be returned here and it's defined / accessible to this test file, import it and use it here)

src/auth/__mocks__/auth.controller.ts
```diff
@@ -0,0 +1,17 @@
+import { faker } from '@faker-js/faker';
+
+export class mockAuthController {
```
rad20c, line 3: class should be in `PascalCase` so:
```suggestion
export class MockAuthController {
```

src/auth/__mocks__/auth.controller.ts
```diff
@@ -0,0 +1,17 @@
...
+  public signIn(_request: { email: string; password: string }) {
```
rad20c, line 4: I think you should be as close to the original method as possible:
```suggestion
  signIn(signInRequest: SignInRequest): Promise<string> {
```

I know I suggested `public` in slack conversation but I think you can drop it here if in the original method youd on't have it either

src/auth/__mocks__/auth.controller.ts
```diff
@@ -0,0 +1,17 @@
...
+    return Promise.resolve({
+      accessToken: faker.string.sample(64),
+    });
+  }
+
+  public signUp(request: { email: string; password: string }) {
```
rad20c, line 10: as above

src/auth/auth.controller.spec.ts
```diff
@@ -1,18 +1,59 @@
...
-  let controller: AuthController;
+  let authController: AuthController;
 
   beforeEach(async () => {
     const module: TestingModule = await Test.createTestingModule({
       controllers: [AuthController],
+      providers: [
+        {
+          provide: AuthService,
+          useClass: mockAuthController,
```
rad20c, line 16: you should inject here `MockAuthService`, not `MockAuthController`

src/auth/auth.service.spec.ts
```diff
@@ -1,18 +1,123 @@
...
+      const request: MockRequest = {
+        email: faker.internet.email(),
+        password: faker.internet.password(),
+      };
+
+      jest.spyOn(userService, 'createUser').mockImplementation((request) => {
+        return Promise.resolve({
+          id: faker.number.int(),
+          email: request.email,
+          password: request.password,
```
rad20c, line 75: didn't we fix it already? We don't want to include user password in the response

src/auth/auth.service.spec.ts
```diff
@@ -1,18 +1,123 @@
...
+      });
+
+      //when
+      const signedUpUser = await authService.signUp(request);
+
+      //then
+      expect(signedUpUser).toEqual({
+        id: expect.any(Number),
+        email: expect.any(String),
+        password: expect.any(String),
```
rad20c, line 87: same as above

src/auth/auth.service.spec.ts
```diff
@@ -1,18 +1,123 @@
...
+      jest.spyOn(userService, 'findOneUser').mockImplementation(() => result);
+
+      //when
+      const validatedUser = await authService.validateUser(request);
+
+      //then
+      expect(validatedUser).toEqual({
+        id: expect.any(Number),
+        email: expect.any(String),
+        password: expect.any(String),
```
rad20c, line 118: as above

src/auth/auth.service.spec.ts
```diff
@@ -1,18 +1,123 @@
...
+
+  describe('ValidateUser', () => {
+    it('should validate user', async () => {
+      //given
+      const request: MockRequest = {
+        email: faker.internet.email(),
+        password: 'john cena',
+      };
+
+      const hashed_password = await bcrypt.hash(request.password, 7);
```
rad20c, line 101: what is 7? If this is SALT_ROUNDS then it should be dependent on the environment variables (you can set different ones for test environment)

src/auth/auth.service.spec.ts
```diff
@@ -1,18 +1,123 @@
...
+      const result = Promise.resolve({
+        id: faker.number.int(),
+        email: faker.internet.email(),
+        password: hashed_password,
+        role: Role.USER,
+      });
+
+      jest.spyOn(userService, 'findOneUser').mockImplementation(() => result);
```
rad20c, line 109: Not a mistake, but I would put `result` contents in `mockImplementation` function as `result` is not used anywhere in this test and `result` variable could be mistaken as the test main result

src/recipe/__mocks__/recipe.controller.ts
```diff
@@ -0,0 +1,63 @@
+import { faker } from '@faker-js/faker';
+class mockRecipeRequest {
+  title: string;
+  description: string;
+  ingredients: string;
+  preparation: string;
+}
+
+export class mockRecipeController {
+  public createRecipe(dto: mockRecipeRequest) {
```
rad20c, line 10: this should use `createRecipeRequest: CreateRecipeRequest`

src/recipe/__mocks__/recipe.controller.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+  public updateRecipe(id: number, dto: mockRecipeRequest) {
+    return Promise.resolve({
+      id,
+      createdAt: Date.now(),
+      title: dto.title || faker.word.noun(),
+      description: dto.description || faker.lorem.text(),
+      ingredients: dto.ingredients || faker.lorem.words(4),
+      preparation: dto.preparation || faker.lorem.lines(5),
+    });
```
rad20c, line 57: this should use `payload: UpdateRecipeRequest` (though in the controller I would name it `updateRecipeRequest: UpdateRecipeRequest` and do the same here)

If you are worried that calls like `description: dto.description || faker.lorem.text(),` might get too long you can do something like this:

```suggestion
  public updateRecipe(id: number, updateRecipeRequest: UpdateRecipeRequest = {}) {
    const { title, description, ingredients, preparation } = updateRecipeRequest;
    
    return Promise.resolve({
      id,
      createdAt: Date.now(),
      title: title || faker.word.noun(),
      description: description || faker.lorem.text(),
      ingredients: ingredients || faker.lorem.words(4),
      preparation: preparation || faker.lorem.lines(5),
    });
```

rad20c: also you should preferably use `??` instead of `||`. Read [here](https://stackoverflow.com/questions/61480993/when-should-i-use-nullish-coalescing-vs-logical-or) on a reason why.

src/auth/auth.service.spec.ts
KubaBaniak: Shouldn't service return password here, so controller can map it and 'ignore' it?

rad20c: If controller does not need to return a `password` field in a response, then there is no need to provide it to the controller. If `password` is not provided, controller won't have to ignore it, because there is nothing to ignore.

Service could return a `password` field if there was some other logic using this service method requiring `password` field for something, but I don't believe we have such logic here.

rad20c (approved):

PR #15, opened 2023-06-30
test: e2e testing

KubaBaniak:
## [Write e2e tests](https://app.clickup.com/t/862k0ag5b)


test/auth.e2e-spec.ts
```diff
@@ -0,0 +1,86 @@
...
+  });
+
+  describe('/auth/signup (POST)', () => {
+    it('should register a user and return the new user object', async () => {
+      request(app.getHttpServer())
+        .post('/auth/signup')
+        .set('Accept', 'application/json')
+        .send(mockUserRequest)
+        .expect(HttpStatus.FORBIDDEN);
+    });
```
rad20c, line 35: something is not right in this test - you say it `should register a user and return the new user object` but it excepts 403 Forbidden status and not checking response body if the user object exists.

test/auth.e2e-spec.ts
```diff
@@ -0,0 +1,86 @@
...
+        .send(mockUserRequest)
+        .expect((response: request.Response) => {
+          const { id, email, role } = response.body;
+          expect(id).toEqual(expect.any(Number));
+          expect(email).toEqual(mockUserRequest.email);
+          expect(role).toEqual(Role.USER);
+        })
+        .expect(HttpStatus.FORBIDDEN);
+    });
+  });
```
rad20c, line 49: I cannot find one important piece - how do we create this user in db (where)?

test/auth.e2e-spec.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    it('should not generate access token for user (wrong email) and return 401 error (UNAUTHENTICATED)', async () => {
+      request(app.getHttpServer())
+        .post('/auth/signin')
+        .set('Accept', 'application/json')
+        .send({
+          email: mockUserRequest.email,
+          password: mockUserRequest.password + 'iLoveYouDad',
+        })
+        .expect(HttpStatus.OK);
```
rad20c, line 68: test descriptions says it expects 401 Unauthorized error but returns 200 OK

test/auth.e2e-spec.ts
```diff
@@ -0,0 +1,86 @@
...
+    });
+
+    it('should not generate access token for user (wrong email) and return 401 error (UNAUTHENTICATED)', async () => {
+      request(app.getHttpServer())
+        .post('/auth/signin')
+        .set('Accept', 'application/json')
+        .send({
+          email: mockUserRequest.email,
+          password: mockUserRequest.password + 'iLoveYouDad',
```
rad20c, line 66: these are some nice words! ❤️ but if you remember what I was saying about the developers wanting to put easter eggs here and there then you will also remember that I prefer e.g. using faker for generating random characters we need 😉

test/get.jwt.token.ts
```diff
@@ -0,0 +1,19 @@
+import { INestApplication } from '@nestjs/common';
+import * as request from 'supertest';
+
+export class JwtTokenExtractor {
+  public JwtToken: Promise<string>;
```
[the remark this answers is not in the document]
rad20c: hmm, looking at it again, why is it `Promise<string>`? I think it should be just `string`. In the `getToken` method where we are assigning the value I think `loginResponse.body` should be a `string`?

test/get.jwt.token.ts
```diff
@@ -0,0 +1,19 @@
...
+  public async getToken(app: INestApplication): Promise<string> {
+    const loginResponse = await request(app.getHttpServer())
+      .post('/auth/signin')
+      .set('Accept', 'application/json')
+      .send({
+        email: 'test@gmail.com',
+        password: 'testpassword123',
+      });
+    this.JwtToken = loginResponse.body;
+    console.log(loginResponse.body);
```
rad20c, line 16: remove console.log

test/get.jwt.token.ts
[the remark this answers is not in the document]
rad20c: looking at it yet again, I think it should be `private`, and in `getToken` method we should first check if it exists and if so return it, otherwise make a request

rad20c (changes requested):
see my comments

test/auth.e2e-spec.ts
KubaBaniak: i defined this user in mockUserRequest after describing `AuthController (e2e)` test, registered it in test you commented above and then in this example i try to add to db same user from `mockUserRequest`.

rad20c: ahh ok now I get it how it works - we need to discuss it 😉

test/mock.recipe.ts
```diff
@@ -0,0 +1,8 @@
+import { faker } from '@faker-js/faker';
+
+export const recipeData = {
```
rad20c, line 3: would be good to use existing type for it - maybe type from Prisma would match? (omitting `id`)

rad20c: Consider creating a factory method that will create a different recipe each time it's called, e.g.:

```typescript
export const createRecipe = (overrides = {}) => ({
  title: overrides.title ?? faker.word.noun(5),
  description: overrides.description ?? faker.lorem.text(),
  ingredients: overrides.ingredients ?? faker.lorem.word(8),
  preparation: overrides.preparation ?? faker.lorem.lines(4),
})
```
or:

```typescript
export const createRecipe = (overrides = {}) => {
  const { title, description, ingredients, preparation } = overrides;

return {
    title: overrides.title ?? faker.word.noun(5),
    description: overrides.description ?? faker.lorem.text(),
    ingredients: overrides.ingredients ?? faker.lorem.word(8),
    preparation: overrides.preparation ?? faker.lorem.lines(4),
  }
}
```

or if you want you can even implement a builder pattern

test/mock.user.ts
```diff
@@ -0,0 +1,6 @@
+import { faker } from '@faker-js/faker';
+
+export const userData = {
+  email: faker.internet.email(),
+  password: faker.internet.password(),
+};
```
rad20c, line 6: as above

test/auth.e2e-spec.ts
```diff
@@ -0,0 +1,99 @@
...
+    prismaService = moduleRef.get<PrismaService>(PrismaService);
+    authService = moduleRef.get<AuthService>(AuthService);
+    await app.init();
+  });
+
+  afterEach(async () => {
+    await prismaService.user.deleteMany();
+  });
+
+  describe('/auth/signup (POST)', () => {
```
rad20c, line 33: hmm, I would perhaps do `POST /auth/signup`, but it's just me preference though

test/mock.recipe.ts
KubaBaniak: Unfortunately, your code gives me error's `Property <...> does not exist on type {}`
What about this? 
```ts
type createRecipeInput = {
  title?: string;
  description?: string;
  ingredients?: string;
  preparation?: string;
};
export const createRecipe = (overrides: createRecipeInput = {}) => ({
  title: overrides.title ?? faker.word.noun(5),
  description: overrides.description ?? faker.lorem.text(),
  ingredients: overrides.ingredients ?? faker.lorem.word(8),
  preparation: overrides.preparation ?? faker.lorem.lines(4),
});
```

KubaBaniak: I used my own type. Prisma has types such `RecipeUpdateInput` and `RecipeCreateInput`, but nothing in between, which I need to pass it to both `Update` and `Create`. I think it would be weird if I pass request with type `RecipeUpdateInput` to create function and vice versa.

rad20c: I wrote it by hand - I haven't tested it nor included everything there. Yes, interface or type is required, but it should be `PascalCase` not `camelCase`, so `CreateRecipeInput`. Also I don't really like `CreateRecipeInput` name, as it's not input, it's overrides. Maybe `CreateRecipeOverrides`?

test/recipe.e2e-spec.ts
```diff
@@ -0,0 +1,234 @@
...
+    it('should not delete recipe and return 404 error (NOT FOUND)', async () => {
+      const recipe = await prismaService.recipe.create({ data: recipeData });
+      return request(app.getHttpServer())
+        .delete(`/recipes/${recipe.id + 1}`)
+        .set({ Authorization: `Bearer ${accessToken}` })
+        .expect(HttpStatus.NOT_FOUND);
+    });
+
+    it('should not delete specific recipe and return 401 error (UNAUTHORIZED)', async () => {
+      const recipe = await prismaService.recipe.create({ data: recipeData });
```
rad20c, line 224: as recipe is created in every test within this `describe` block, you can move recipe creation to `beforeEach` - of course within this describe block

rad20c (approved):
good job 👍🏻

PR #16, opened 2023-07-04
task: configure CircleCI

KubaBaniak:
## [Configure CI pipeline](https://app.clickup.com/t/862k1pe9j)


.circleci/config.yml
```diff
@@ -0,0 +1,79 @@
...
+    docker:
+      - image: cimg/base:2023.07
+      - image: cimg/postgres:14.1
+
+    steps:
+      - checkout
+      - install_dependencies
+      - setup_prisma
+      - setup_docker_compose
+      - run: echo "BUILDING PROJECT"
```
arkus7, line 68: I guess this is a placeholder?

.circleci/config.yml
```diff
@@ -0,0 +1,79 @@
...
+      - run:
+          name: Run unit tests
+          command: npm run test
+
+  deploy:
+    working_directory: ~/repo
+
+    docker:
+      - image: cimg/base:2023.07
+      - image: cimg/postgres:14.1
```
arkus7, line 61: Why do you need postgres for deployment?

arkus7 (changes requested):
Use the `cimg/node` image instead of `base`

.circleci/config.yml
KubaBaniak: Thought, I have to deploy the app as well as db for it. With that suggestion I assume db will be hosted on different server?

arkus7 (approved):

PR #17, opened 2023-07-04
task: configure Husky

KubaBaniak:
### Acceptance criteria:
- Husky is configured in the project and automatically installed on project setup
- When committing the code:
    - we should validate if commit names follow the conventional commit naming pattern (check out commitlint ) 
    - we should lint the code using eslint and/or prettier (check out lint-staged )


src/app.controller.spec.ts
```diff
@@ -15,6 +15,8 @@ describe('AppController', () => {
   });
 
   describe('root', () => {
-    it('should return "Hello World!"', () => {});
+    it('should return "Hello World!"', () => {
```
wojdanowski, line 18: this looks like something that you do not need :)

package.json
```diff
@@ -68,7 +76,8 @@
     "ts-loader": "9.2.3",
     "ts-node": "10.9.1",
     "tsconfig-paths": "4.2.0",
-    "typescript": "4.9.5"
+    "typescript": "4.9.5",
+    "husky": "^8.0.0"
```
wojdanowski, line 80:
```suggestion
    "husky": "8.0.0"
```

I would be strict with versioning. 
I would add 
`save-exact=true` to `.npmrc`

wojdanowski (approved):

PR #21, opened 2023-07-06
chore: add separate job to build and test

KubaBaniak:
### Acceptance criteria:
- we have a workflow divided into multiple jobs
- there is a job responsible for repository checkout and installing the dependencies
- there is a job responsible for running unit tests
- there is a job responsible for running e2e tests
- unit and e2e tests jobs are running in parallel (at the same time)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,22 +25,55 @@ commands:
           command: docker-compose up -d
 
 jobs:
-  build_and_test:
+  build:
     working_directory: ~/repo
     docker:
       - image: cimg/node:18.16.0
       - image: cimg/postgres:14.1
+    resource_class: medium
     steps:
       - checkout
+      - restore_cache:
+          key: v1-deps-{{ checksum "package-lock.json" }}
       - install_dependencies
+      - save_cache:
+          paths:
+            - './node_modules'
+          key: v1-deps-{{ checksum "package-lock.json" }}
+
+  unit_test:
+    docker:
+      - image: cimg/node:18.16.0
+      - image: cimg/postgres:14.1
+    parallelism: 2
+    resource_class: medium
+    steps:
+      - checkout
+      - restore_cache:
+          key: v1-deps-{{ checksum "package-lock.json" }}
       - setup_prisma
       - setup_docker_compose
       - run:
           name: Run unit tests
-          command: npm run test
+          command: |
+            npm test -- $(circleci tests glob "src/*/*{spec}.ts" | circleci tests split --split-by=timings)
+
+  e2e_test:
+    docker:
+      - image: cimg/node:18.16.0
+      - image: cimg/postgres:14.1
+    resource_class: medium
+    parallelism: 2
+    steps:
+      - checkout
+      - restore_cache:
+          key: v1-deps-{{ checksum "package-lock.json" }}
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -25,21 +25,48 @@ commands:
...
+  build:
     working_directory: ~/repo
     docker:
       - image: cimg/node:18.16.0
       - image: cimg/postgres:14.1
+    resource_class: medium
     steps:
       - checkout
+      - restore_cache:
+          key: v1-deps-{{ checksum "package-lock.json" }}
```
rad20c, line 37: I recommend reading `v1` from environment variables - it will give you control over cache without deploying changes in the code, e.g.:
```suggestion
          key: {{ .Environment.CACHE_VERSION }}-deps-{{ checksum "package-lock.json" }}
```

.circleci/config.yml
```diff
@@ -25,21 +25,48 @@ commands:
           command: docker-compose up -d
 
 jobs:
-  build_and_test:
+  build:
     working_directory: ~/repo
     docker:
       - image: cimg/node:18.16.0
       - image: cimg/postgres:14.1
```
rad20c, line 32: at the top of this file you can add something like this:

```yaml
defaults: &defaults
  working_directory: ~/repo
  docker:
    - image: cimg/node:18.16.0
    - image: cimg/postgres:14.1
```

and here do:
```suggestion
    <<: *defaults
```

This way you don't have to repeat this part every time. If you need to use other images or working directory you just define the key the same way as right now - it will override given key from `defaults`

.circleci/config.yml
```diff
@@ -25,21 +25,48 @@ commands:
...
+      - image: cimg/postgres:14.1
+    steps:
+      - checkout
+      - restore_cache:
+          key: v1-deps-{{ checksum "package-lock.json" }}
+      - setup_prisma
+      - setup_docker_compose
       - run:
-          name: Run e2e tests
+          name: Run unit tests
```
rad20c, line 69: This should be `Run e2e tests` right?

.circleci/config.yml
```diff
@@ -25,21 +25,48 @@ commands:
...
       - install_dependencies
+      - save_cache:
+          paths:
+            - './node_modules'
+          key: v1-deps-{{ checksum "package-lock.json" }}
+
+  unit_test:
+    docker:
+      - image: cimg/node:18.16.0
+      - image: cimg/postgres:14.1
```
rad20c, line 47: missing `resource_class`

.circleci/config.yml
```diff
@@ -50,12 +77,19 @@ jobs:
       - run: echo "!TODO"
 
 workflows:
-  build_and_test:
+  Recipe_app_workflow:
```
rad20c, line 80: hmm, not sure about the name change here. I'd leave the previous one I guess. If you plan to have 1 workflow only then I guess this name is ok, but personally I would go for a separate workflow for the deployment

rad20c (approved):

.circleci/config.yml
rad20c: of course this is optional - some devs don't like using `defaults` as for them it's not a readable as it could be

PR #23, opened 2023-07-10
feat: implement cache

KubaBaniak:
### Acceptance criteria:
- `GET` endpoints that are returning recipe related data are using cache
- When modifying any data related to the recipes the appropriate cache is deleted


docker-compose.yml
```diff
@@ -24,9 +24,22 @@ services:
       - tools
     networks:
       - recipe-app-network
+  cache:
+    image: redis:7.0
```
rad20c, line 28: I'd use a more specific version - `redis:7.0.12`. Going even further - I'd try a `redis:7.0.12-apline` first. Alpine is a super lightweight linux distribution, but because of this some feature might not work - that's why we need to test it before going with it.

example.env
```diff
@@ -6,3 +6,4 @@ PGADMIN_PASSWORD=
 JWT_SECRET=
 BCRYPT_SALT_ROUNDS=
 BCRYPT_SALT_TEST=
+REQUIRE_PASS_REDIS=
```
rad20c, line 9: hmm, kind of strange name - wouldn't just `REDIS_PASSWORD` be better?

docker-compose.yml
```diff
@@ -24,9 +24,22 @@ services:
...
+    restart: always
+    ports:
+      - '6379:6379'
+    command: redis-server --save 20 1 --loglevel warning --requirepass '${REQUIRE_PASS_REDIS}'
```
rad20c, line 32: hmm, I don't think we need to persist the database for our purposes (remember this is cache only), but it can stay :)

src/cache/redisCache.module.ts
```diff
@@ -0,0 +1,21 @@
...
+import * as redisStore from 'cache-manager-redis-store';
+import { RedisCacheService } from './redisCache.service';
+
+@Module({
+  imports: [
+    CacheModule.register({
+      imports: [ConfigModule],
+      useFactory: async () => ({
+        store: redisStore,
+        url: 'redis://localhost:6379',
```
rad20c, line 13: `url` should be moved to environment variables as it might change with the environment

src/cache/redisCache.service.ts
```diff
@@ -0,0 +1,20 @@
...
+export class RedisCacheService {
+  constructor(@Inject(CACHE_MANAGER) private readonly cache: Cache) {}
+
+  async get(key: string): Promise<any> {
+    return await this.cache.get(key);
+  }
+
+  async set(key: string, value: any) {
+    await this.cache.set(key, value);
+  }
```
rad20c, line 15: I believe you should use generics `<T>` here instead of `any`

rad20c: missing return type - if it returns nothing then set `void` / `Promise<void>`

arkus7, line 15: Consider using generics to specify the types

```suggestion
  async get<T>(key: string): Promise<T> {
    return await this.cache.get(key);
  }

  async set<T>(key: string, value: T) {
    await this.cache.set(key, value);
  }
```

this way you can use type inference when calling your methods.

Example: [Playground](https://www.typescriptlang.org/play?#code/MYGwhgzhAEDCbABYFMDKyBOA3AlsZ0A3gFDTTAD2AdhAC4YCuwtFGAFAA4Y5Zi0EZkYACbUQAT3IIUALmhgq4gJREAvsVLyI4qsGgBzZLQA8AFQB8bANbJxcutyr6lcgAoYKAWxwRkZ80SaZIK0DBhU8gDuYDi00LSIPgB0wNLISYa01rZKANya6pqQOnq+JhbZdtAOOE4ANNC8IAzIcqYqJGRkYNGx8YkQKWlJZZUNTS15BcSFlDRxqUhomLj40AC80FTIkXBp6Nh4yGyEqlPExbrQAGYMurQ41PHIdGwdGmQgRtVeyACqvgwAElhPZ6LV9PlNHM6I0wM0CJsejEFvsVkcMkY2AByBiAkHYpSaAD0xK60AAegB+D5whFJFgAGQokUw8F8byhZBhcTxmBBGyiKKkSwOq3SmWMNScllx+OEhPyZFJ5OptL5wOEDIozNZGHZx3O3OosI1IIATGDHPpBci+osUGKMZkcWaFVNlWSumrNBBfgD+cJbb1UaL0fhMVlsRQEpgA5rFSSyWrVEA)

You shouldn't use `any` as a type, prefer using `unknown`

src/cache/redisCache.service.ts
```diff
@@ -0,0 +1,20 @@
+import { Injectable, Inject } from '@nestjs/common';
+import { Cache } from 'cache-manager';
+import { CACHE_MANAGER } from '@nestjs/cache-manager';
+
+@Injectable()
+export class RedisCacheService {
+  constructor(@Inject(CACHE_MANAGER) private readonly cache: Cache) {}
+
+  async get(key: string): Promise<any> {
+    return await this.cache.get(key);
```
arkus7, line 10: Redundant `await` if `this.cache.get` returns a Promise already
```suggestion
    return this.cache.get(key);
```

src/cache/redisCache.service.ts
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  async set(key: string, value: any) {
```
arkus7, line 13: Missing typedef for a return value. It's really easy to introduce errors when you don't specify the types of returned values, [as TS will infer the type from the function](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAJwKZRMsBnAKqgDygAoBKRAbwChEV1MkByAQ2wBNGBuKgXyqtCRYCOhix5CUAExlKNUQ0qIokgFyIW7DT258qEBNiiIAbswA2ARkQBeBePxEy3APQvatAHoB+fgZzGZuZStvY4jtLOVG4eiD78QZYAdFBwADJwAO6oyADCrKhRQVIp6Vk5+diFpJxAA) (it's not always the type you actually want to be returned).

```suggestion
  async set(key: string, value: any): Promise<void> {
```

src/cache/redisCache.service.ts
```diff
@@ -0,0 +1,20 @@
...
+    await this.cache.set(key, value);
+  }
+
+  async del(key: string) {
```
arkus7, line 17: missing typedef

src/recipe/recipe.controller.spec.ts
```diff
@@ -1,22 +1,30 @@
 import { Test, TestingModule } from '@nestjs/testing';
 import { RecipeService } from './recipe.service';
 import { RecipeController } from './recipe.controller';
-import { PrismaService } from 'src/prisma/prisma.service';
+import { PrismaService } from '../prisma/prisma.service';
 import { faker } from '@faker-js/faker';
 import { MockRecipeService } from './__mocks__/recipe.service.mock';
+import { MockRedisCacheService } from '../cache/__mock__/redis.cache.mock';
+import { RedisCacheModule } from '../cache/redisCache.module';
+import { RedisCacheService } from 'src/cache/redisCache.service';
```
arkus7, line 9: absolute import
```suggestion
import { RedisCacheService } from '../cache/redisCache.service';
```

src/recipe/recipe.controller.ts
```diff
@@ -65,9 +75,13 @@ export class RecipeController {
...
   ): Promise<FetchRecipeResponse> {
-    const fetchedRecipe = await this.recipeService.fetchRecipe(id);
+    let recipe: Recipe = await this.redisCacheService.get(`${id}`);
 
-    return FetchRecipeResponse.from(fetchedRecipe);
+    if (!recipe) {
+      recipe = await this.recipeService.fetchRecipe(id);
+      await this.redisCacheService.set(`cache:recipe:${recipe.id}`, recipe);
+    }
+    return FetchRecipeResponse.from(recipe);
```
rad20c, line 84: yep 👍🏻 but see my comment above - this should be in a service

src/recipe/recipe.controller.ts
```diff
@@ -48,6 +53,11 @@ export class RecipeController {
     const createdRecipe = await this.recipeService.createRecipe(
       createRecipeRequest,
     );
+    await this.redisCacheService.set(
+      `cache:recipe:${createdRecipe.id}`,
```
arkus7, line 57: Consider extracting a function for creating a key for a cached recipe.
It's super easy to do a typo in the keys (as they're usually strings) and super hard to find such an issue.

```ts
recipeCacheKey(recipeId: string): string {
  return `cache:recipe:${createdRecipe.id}`;
}
```
and use it like this
```suggestion
      recipeCacheKey(createdRecipe.id),
```

src/recipe/recipe.controller.ts
```diff
@@ -116,5 +132,6 @@ export class RecipeController {
   @Delete(':id')
   async deleteRecipe(@Param('id', ParseIntPipe) id: number): Promise<void> {
     await this.recipeService.deleteRecipe(id);
+    await this.redisCacheService.del(`cache:recipe:${id}`);
```
rad20c, line 135: This is the 4th time we are repeating the same string (cache key) - that's why I recommend moving it to a dedicated cache service and perhaps creating a private method that will return it

src/recipe/recipe.controller.ts
```diff
@@ -65,9 +75,13 @@ export class RecipeController {
   async fetchRecipe(
     @Param('id', ParseIntPipe) id: number,
   ): Promise<FetchRecipeResponse> {
-    const fetchedRecipe = await this.recipeService.fetchRecipe(id);
+    let recipe: Recipe = await this.redisCacheService.get(`${id}`);
```
arkus7, line 78: Well, just like mentioned above, you're storing the recipe in the cache under `cache:recipe:${createdRecipe.id}` key, but try to read it from `${id}` key.

Use the `recipeCacheKey` mentioned above in all places.

src/recipe/recipe.controller.ts
```diff
@@ -65,9 +75,13 @@ export class RecipeController {
...
 
-    return FetchRecipeResponse.from(fetchedRecipe);
+    if (!recipe) {
+      recipe = await this.recipeService.fetchRecipe(id);
+      await this.redisCacheService.set(`cache:recipe:${recipe.id}`, recipe);
```
arkus7, line 82: As above

src/recipe/recipe.controller.ts
arkus7: You could also create a special type using [Template Literal Types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) to use the type system to guard you against such issues.

[Playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAShDGBLSBhAhvAFhA0hEAPHEpAJIAmUEAHsBAHbkDOUTwATovQOZQC8rDl24A+flAAG8DNgBc7BMgiyAJAG9iSigF8JAbgBQBgGYBXevGCIA9vSjcIwTZABi7awFt0WCAAoA1viysIqoMrj4AJTB5v701gDudmoGUFAKwKbsydqG2kYOTqEQbp7e2L4A5NI+8sWyaEzklZGGhc4l7l7hVY3NrUA)

This way, you also won't be able to e.g. overwrite recipe data with user data, because the key won't have proper structure.

src/recipe/recipe.controller.ts
```diff
@@ -96,12 +110,14 @@ export class RecipeController {
...
   ): Promise<UpdatedRecipeResponse> {
-    const updatedRecipe = await this.recipeService.updateRecipe(
-      id,
-      updateRecipeRequest,
-    );
+    let recipe: Recipe = await this.redisCacheService.get(`${id}`);
+
+    if (!recipe) {
+      recipe = await this.recipeService.updateRecipe(id, updateRecipeRequest);
+      await this.redisCacheService.del(`cache:recipe:${recipe.id}`);
```
arkus7, line 117: As above, use a method for the key to avoid issues with typos

src/recipe/recipe.controller.ts
```diff
@@ -96,12 +110,14 @@ export class RecipeController {
...
+    }
```
rad20c, line 118: not sure what's happening here - why do we need to fetch a recipe from cache to update it?

src/recipe/recipe.controller.ts
```diff
@@ -96,12 +110,14 @@ export class RecipeController {
     @Param('id', ParseIntPipe) id: number,
     @Body() updateRecipeRequest: UpdateRecipeRequest,
   ): Promise<UpdatedRecipeResponse> {
-    const updatedRecipe = await this.recipeService.updateRecipe(
-      id,
-      updateRecipeRequest,
-    );
+    let recipe: Recipe = await this.redisCacheService.get(`${id}`);
```
arkus7, line 113: Same issue as above

src/recipe/recipe.controller.ts
arkus7, line 135: as above

arkus7 (changes requested):

rad20c (changes requested):
see my comments

example.env
KubaBaniak: I used it because if I have worked on local Redis, I would have to pass `requirepass` in `redis.conf`. And in docs they always define this password with such keyword so i sticked with that.

KubaBaniak: Oh, looking at the next comment, classic REDIS_PASSWORD is much more pleasing to the eye

arkus7: `--requirepass` is an argument that can be read as `Require password? If so, what should be the value?`

It's easier to understand what's the meaning of the value, when the env variable name is `REDIS_PASSWORD` than `REQUIRE_PASS_REDIS` 😉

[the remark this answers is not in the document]
KubaBaniak: To set password for redis, `requirepass <password>` is needed. Thats why I picked such name. But yea - passing `REDIS_PASSWORD` to `docker-compose` is way more clear

prisma/migrations/20230710111854_/migration.sql
```diff
@@ -0,0 +1,8 @@
+/*
+  Warnings:
+
+  - Made the column `description` on table `Recipe` required. This step will fail if there are existing NULL values in that column.
```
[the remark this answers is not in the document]
KubaBaniak: No, I do not recall that I switched description to optional. In request validations i set description to optional as well, so I don't know why thats here tbh.

src/cache/redisCache.module.ts
rad20c, line 1: haven't noticed previously, but to be consistent with the naming convention we are following, this file should be named `redis-cache.module.ts`

src/recipe/__mocks__/recipe.cache.mock.ts
```diff
@@ -0,0 +1,16 @@
+import { RecipeCacheService } from '../recipe.cache.service';
+import { Recipe } from '@prisma/client';
+
+export class MockRecipeCacheService implements Required<RecipeCacheService> {
+  getCachedRecipe(_id: number): Promise<any> {
```
rad20c, line 5: replace `any` with a proper type

src/recipe/recipe.cache.service.ts
```diff
@@ -0,0 +1,24 @@
+import { Injectable } from '@nestjs/common';
+import { RedisCacheService } from '../cache/redisCache.service';
+import { Recipe } from '@prisma/client';
+
+function recipeCacheKey(recipeId: number): string {
+  return `cache:recipe:${recipeId}`;
+}
```
rad20c, line 7: hmm, why is this function outside of class? You can define it as a private method in `RecipeCacheService` class - it won't be publicly available (and I don't believe it there should be a need for it being public)

rad20c (approved):

src/recipe/recipe.cache.service.ts
KubaBaniak: wanted to call `recipeCacheKey(...)` rather than `this.recipeCacheKey(...)`

arkus7 (approved):

PR #24, opened 2023-07-13
feat: show recipes based on ownership and visibility

KubaBaniak:
### Acceptance criteria:
- The recipes can now be private or public.
- User can see all of the recipes that he is the owner of.
- User cannot see private recipes created by other users.
- User can see public recipes created by other users.
- Admin can see / edit / delete all recipes in the system.
- Besides admin, only the owner can edit / delete the recipe.
- Recipe visibility can be set when creating recipe and also when updating it.


src/prisma/__mocks__/prisma.service.mock.ts
```diff
@@ -31,6 +33,16 @@ export class MockPrismaService {
...
       return Promise.resolve();
     },
+
+    findUnique: function (request: { where: { id: number } }): Promise<User> {
+      userId = request.where.id;
+      return Promise.resolve({
+        id: userId,
+        email: faker.internet.email(),
+        password: faker.internet.password(),
+        role: 'USER',
```
PrzerostNadSens, line 43:
```suggestion
        role: Role.USER,
```
always use predefined variables if possible
(this will improve the readability of the code and make future refactoring easier)

src/prisma/__mocks__/prisma.service.mock.ts
```diff
@@ -2,6 +2,8 @@ import { faker } from '@faker-js/faker';
 import { Prisma, Recipe, User } from '@prisma/client';
```
PrzerostNadSens, line 2:
```suggestion
import { Prisma, Recipe, Role, User } from '@prisma/client';
```

prisma/schema.prisma
```diff
@@ -24,6 +25,9 @@ model Recipe {
   description String?
   ingredients String
   preparation String
+  isPublic      Boolean  @default(false)
```
wojdanowski, line 28: too many tabs?

src/recipe/__mocks__/recipe.controller.mock.ts
```diff
@@ -16,25 +16,63 @@ export class MockRecipeController {
...
+      {
+        id: 1,
+        createdAt: new Date(),
+        title: faker.word.noun(),
+        description: faker.lorem.text(),
+        ingredients: faker.lorem.words(4),
+        preparation: faker.lorem.lines(5),
+        isPublic: true,
+        authorId,
+      },
```
wojdanowski, line 69: I would add some kind of method to generate recipes for mocks and tests :
something like: generateRecipeResponse(): RecipeResponse

src/recipe/__mocks__/recipe.service.mock.ts
```diff
@@ -3,26 +3,61 @@ import { CreateRecipeRequest, UpdateRecipeRequest } from '../dto';
...
   }
 
-  fetchAllRecipes(): Promise<Recipe[]> {
+  fetchUsersRecipes(
+    _req: { userId: number },
+    _email: string,
+  ): Promise<Recipe[]> {
+    const authorId = faker.number.int();
+    return Promise.all([
+      {
```
wojdanowski, line 37: same as above for generateRecipeResponse but for users

src/recipe/dto/update-recipe-request.ts
```diff
@@ -14,4 +14,7 @@ export class UpdateRecipeRequest {
 
   @IsOptional()
   preparation: string;
+
+  @IsOptional()
+  isPublic: boolean;
```
wojdanowski, line 19: if property is optional is should also be reflected in the typing:
```suggestion
  isPublic?: boolean;
```

src/recipe/recipe.controller.spec.ts
```diff
@@ -38,48 +38,100 @@ describe('RecipeController', () => {
...
           title: expect.any(String),
           description: expect.any(String),
           ingredients: expect.any(String),
           preparation: expect.any(String),
+          isPublic: true,
+          authorId: userObjectId.user.id,
         },
       });
     });
+    it('should fetch all users recipe recipes', async () => {
```
wojdanowski, line 96:
```suggestion
    it('should fetch all users recipes', async () => {
```

src/recipe/recipe.controller.ts
```diff
@@ -64,19 +67,50 @@ export class RecipeController {
...
+  @UseGuards(JwtAuthGuard)
+  @ApiOperation({ summary: 'Get all users recipes' })
+  @ApiBearerAuth()
+  @ApiUnauthorizedResponse({ description: 'User is not authenticated' })
+  @ApiParam({
+    name: 'email',
+    description: 'Email of a user',
+  })
+  @Get('/user/:email')
+  async fetchUsersRecipes(
```
wojdanowski, line 91: `fetchUsersRecipes` is a bit confusing in my opinion. 
how about 
`findRecipesByAuthorEmail()` or ...OwnerEmail

src/recipe/recipe.controller.ts
```diff
@@ -64,19 +67,50 @@ export class RecipeController {
...
+  }
+
   @UseGuards(JwtAuthGuard)
   @ApiOperation({ summary: 'Get list of all recipes' })
   @ApiBearerAuth()
   @ApiUnauthorizedResponse({ description: 'User is not authenticated' })
   @Get()
-  async fetchRecipes(): Promise<FetchRecipesResponse> {
-    const fetchedRecipes = await this.recipeService.fetchAllRecipes();
+  async fetchRecipes(
```
wojdanowski, line 108: how about findRecipes() ?

src/recipe/recipe.controller.ts
```diff
@@ -43,10 +44,12 @@ export class RecipeController {
   @ApiUnauthorizedResponse({ description: 'User is unauthorized' })
   @Post()
   async createRecipe(
+    @Request() userIdObject: { user: { id: number } },
```
wojdanowski, line 47: I would add some kind of custom decorator for retrieving authorized user object from the request.
Keep in mind that `@Request` will save the entire request in `userIdObject` while you only need the `userId`

src/recipe/recipe.service.ts
```diff
@@ -35,35 +57,117 @@ export class RecipeService {
...
+    }
+  }
+
+  async fetchAllRecipes(userId: number): Promise<Recipe[]> {
+    const user = await this.prisma.user.findUnique({
+      where: {
+        id: userId,
+      },
+    });
+    if (user.role === 'ADMIN') {
```
wojdanowski, line 108: `ADMIN` in many places looks like something that should be an enum value

src/recipe/recipe.service.ts
```diff
@@ -35,35 +57,117 @@ export class RecipeService {
...
+      });
+    } else {
+      return this.prisma.recipe.findMany({
+        where: {
+          authorId: recipesOwner.id,
+          isPublic: true,
+        },
+      });
+    }
+  }
```
wojdanowski, line 100: You should think about some way to avoid having that many if statements.

Service that is responsible for making calls to DB should not contain that kind of logic
`recipesOwner.id === visitor.id || visitor.role == 'ADMIN'` 
or:
`      user.id != recipe.authorId &&
      user.role != 'ADMIN' &&
      !recipe.isPublic`

this looks like a business logic that should be handled on a different level.

test/recipe.e2e-spec.ts
```diff
@@ -132,10 +148,77 @@ describe('RecipeController (e2e)', () => {
...
+          { ...userPrivateRecipe, authorId: user.id },
+          { ...otherPublicRecipe, authorId: otherUser.id },
+          { ...otherPrivateRecipe, authorId: otherUser.id },
+        ],
+      });
+      request(app.getHttpServer())
+        .get(`/recipes`)
+        .set({ Authorization: `Bearer ${accessToken}` })
+        .expect((response: request.Response) => {
+          expect(response.body.fetchedRecipes).toEqual(
```
wojdanowski, line 172: IMO, in e2e tests, you could use mocked service implementation (you do not have to make actual calls to the database)

Obviously the matter of choice.

wojdanowski: Btw, you have a test cases that already test if the response from db is the same as expected. Is there a reason to do it twice?
`                id: expect.any(Number),
                createdAt: expect.any(String),
                title: expect.any(String),
                description: expect.any(String),
                ingredients: expect.any(String),
                preparation: expect.any(String),
                isPublic: expect.any(Boolean),
                authorId: expect.any(user.id),`

KubaBaniak: It's because before sending request, I create 2 recipes ( public and private ) for owner (let's call him that way) and 2 someone else's recipes ( public and private ).  Then the owner wants to get all recipes, so he should get all his as well as all other public recipes. So now in `expect.arrayContaining([...])` we check if it has any public recipe of someone else, and all of his

src/recipe/recipe.service.ts
arkus7, line 108: `Role.Admin`

src/recipe/recipe.service.ts
```diff
@@ -35,35 +57,117 @@ export class RecipeService {
...
+    let recipe = await this.prisma.recipe.findUnique({
+      where: {
+        id: recipeId,
+      },
+    });
+
     try {
-      const recipe = await this.prisma.recipe.update({
-        where: { id },
+      if (user.id != recipe.authorId && user.role != 'ADMIN') {
```
arkus7, line 136: `Role.Admin`

arkus7 (changes requested):

prisma/schema.prisma
arkus7: `npx prisma format`

src/recipe/recipe.controller.ts
KubaBaniak: I think I'm going to stick with this fetch for now. I used `fetch` everywhere in project, so it would require lots of changes (mocks, tests, service, controller, dtos, and likely few things more).  I'll probably create a different refactor branch and do this in it.

prisma/migrations/20230711092234_ownership_and_visibility/migration.sql
```diff
@@ -0,0 +1,14 @@
+/*
+  Warnings:
+
+  - Added the required column `authorId` to the `Recipe` table without a default value. This is not possible if the table is not empty.
```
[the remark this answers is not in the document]
KubaBaniak: No, I create `Recipe` table right after init migration

prisma/migrations/20230711092234_ownership_and_visibility/migration.sql
```diff
@@ -0,0 +1,14 @@
...
+  - Made the column `description` on table `Recipe` required. This step will fail if there are existing NULL values in that column.
```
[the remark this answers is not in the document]
KubaBaniak: To me yes, every recipe should have something but NULL.

src/prisma/__mocks__/prisma.service.mock.ts
```diff
@@ -1,7 +1,9 @@
 import { faker } from '@faker-js/faker';
-import { Prisma, Recipe, User } from '@prisma/client';
+import { Prisma, Recipe, User, Role } from '@prisma/client';
 import { UpdateRecipeRequest } from 'src/recipe/dto';
 
+let userId: number;
```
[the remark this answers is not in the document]
KubaBaniak: Its main purpose is for e2e updateRecipe test. To pass this test, `user id` has to match recipe's `author id`, because either admin or **owner** can update it. Therefore I sticked with this variable, which is being updated every time `user.findUnique` gets called, so the `recipe.findUnique` will mock a recipe with the same `authorId` as user's `id'

src/recipe/recipe.controller.ts
```diff
@@ -64,19 +67,50 @@ export class RecipeController {
...
+  @HttpCode(200)
+  @UseGuards(JwtAuthGuard)
+  @ApiOperation({ summary: 'Get all users recipes' })
+  @ApiBearerAuth()
+  @ApiUnauthorizedResponse({ description: 'User is not authenticated' })
+  @ApiParam({
+    name: 'email',
+    description: 'Email of a user',
+  })
+  @Get('/user/:email')
```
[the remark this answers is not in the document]
KubaBaniak: Thought email tells user a little bit more whose recipes are being fetched. Talked about this with Rafał as well. He suggested to use `id` too.

src/recipe/dto/create-recipe-request.ts
```diff
@@ -21,4 +21,10 @@ export class CreateRecipeRequest {
     example: 'Step 1\nStep 2\nStep 3',
   })
   preparation: string;
+
+  @ApiProperty({
+    description: 'Sets recipe to public',
```
rad20c, line 26: maybe something like this?
```suggestion
    description: 'Sets recipe to be publicly accessible by other users',
```

src/recipe/recipe.controller.ts
```diff
@@ -64,19 +67,46 @@ export class RecipeController {
...
+    @Param('id') authorId: number,
+  ): Promise<FetchRecipesResponse> {
+    const fetchedUsersRecipes = await this.recipeService.fetchRecipesByAuthorId(
+      authorId,
+      userId,
+    );
+
+    return FetchRecipesResponse.from(fetchedUsersRecipes);
+  }
+
```
rad20c, line 102: hmm, I think I wouldn't create a separate endpoint for this. Instead in the existing `GET /recipes` endpoint I would add support for filters in query params, e.g. `GET /recipes?authorId=123456`

src/recipe/recipe.service.ts
```diff
@@ -35,35 +57,123 @@ export class RecipeService {
...
+      }
       await this.prisma.recipe.delete({
-        where: { id },
+        where: { id: recipeId },
       });
-      this.recipeCacheService.deleteCachedRecipe(id);
+      this.recipeCacheService.deleteCachedRecipe(recipeId);
     } catch {
       throw new NotFoundException();
     }
```
rad20c, line 179: I don't understand this part - you are throwing a `ForbiddenException` but it will be catched by `catch` block and `NotFoundException` will be thrown instead. Is there a need for a try catch here at all? I don't think Prisma returns error on missing record?

src/recipe/test/recipe.factory.ts
```diff
@@ -0,0 +1,41 @@
...
+  preparation?: string;
+  isPublic?: boolean;
+};
+
+export const createRecipe = (overrides: CreateRecipeOverrides = {}) => ({
+  title: overrides.title ?? faker.word.noun(5),
+  description: overrides.description ?? faker.lorem.text(),
+  ingredients: overrides.ingredients ?? faker.lorem.word(8),
+  preparation: overrides.preparation ?? faker.lorem.lines(4),
+  isPublic: overrides.isPublic ?? true,
```
rad20c, line 16: should be `false` by default I think (to be consistent with the default `false` value in the database

src/recipe/test/recipe.factory.ts
```diff
@@ -0,0 +1,41 @@
...
+export const createRecipeResponse = (
+  overrides: CreateRecipeResponseOverrides = {},
+) => ({
+  id: overrides.id ?? faker.number.int(),
+  createdAt: overrides.createdAt ?? new Date(),
+  title: overrides.title ?? faker.word.noun(5),
+  description: overrides.description ?? faker.lorem.text(),
+  ingredients: overrides.ingredients ?? faker.lorem.word(8),
+  preparation: overrides.preparation ?? faker.lorem.lines(4),
+  isPublic: overrides.isPublic ?? true,
```
rad20c, line 39: as above

test/auth.e2e-spec.ts
```diff
@@ -23,20 +24,15 @@ describe('AuthController (e2e)', () => {
     app = moduleRef.createNestApplication();
     prismaService = moduleRef.get<PrismaService>(PrismaService);
     authService = moduleRef.get<AuthService>(AuthService);
+    user = createUser();
     await app.init();
   });
 
   afterEach(async () => {
-    await prismaService.user.deleteMany();
+    await prismaService.user.delete({ where: { email: user.email } });
```
rad20c, line 32: not sure about this change

rad20c (changes requested):
check comments

src/recipe/recipe.service.ts
KubaBaniak: `prisma.update` throws an `RecordNotFound` exception if record does not exist ([Prisma Client API reference](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference))

src/recipe/__mocks__/recipe.controller.mock.ts
```diff
@@ -7,74 +6,67 @@ import {
...
   ): Promise<CreateRecipeResponse> {
-    const createdRecipe = {
-      id: faker.number.int(),
-      createdAt: new Date(),
+    const createdRecipe = createRecipeResponse({
+      authorId: userId,
       ...createRecipeRequest,
-    };
+    });
+    console.log(createdRecipe);
```
rad20c, line 20: console.log

src/recipe/recipe.controller.ts
```diff
@@ -75,9 +84,18 @@ export class RecipeController {
...
-  async fetchRecipes(): Promise<FetchRecipesResponse> {
-    const fetchedRecipes = await this.recipeService.fetchAllRecipes();
-
+  async fetchRecipes(
+    @UserId() userId: number,
+    @Query('authorId') authorId?: string,
+  ): Promise<FetchRecipesResponse> {
+    if (authorId) {
+      const fetchedRecipes = await this.recipeService.fetchRecipesByAuthorId(
+        +authorId,
```
rad20c, line 93: check if you can transform it to number via class-validator, so that here you can have `authorId?: number` instead of `string`

src/recipe/recipe.service.ts
KubaBaniak: This has been fixed with repository layer, which separates these things

src/recipe/test/recipe.controller.spec.ts
```diff
@@ -57,29 +63,38 @@ describe('RecipeController', () => {
...
+  describe('FindRecipe', () => {
     it('should fetch recipe given id', async () => {
       //given
-      const id = faker.number.int();
+      const recipeId = faker.number.int();
+      const userId = faker.number.int();
 
       //when
-      const fetchedRecipe = await recipeController.fetchRecipe(id);
+      const fetchedRecipe = await recipeController.fetchRecipe(
```
rad20c, line 73: in describe block you have `find` but in controller it's `fetch` - unify this (for me it doesn't matter which one you'll pick)

rad20c (approved):

PR #25, opened 2023-07-14
feat: implement repository

KubaBaniak:
### Acceptance criteria:
- Each service that has some database operations should have appropriate repository class injected
- All database logic should be moved to appropriate repositories


src/recipe/recipe.service.ts
```diff
@@ -1,20 +1,18 @@
...
 
 @Injectable()
 export class RecipeService {
   constructor(
-    private prisma: PrismaService,
+    private readonly recipeRepository: RecipeRepository,
     private readonly recipeCacheService: RecipeCacheService,
   ) {}
 
   async createRecipe(data: Prisma.RecipeCreateInput): Promise<Recipe> {
```
rad20c, line 14: as per our discussion - perhaps you can get replace `Prisma.RecipeCreateInput` type with some DTO type?

src/user/dto/user-response.ts
```diff
@@ -0,0 +1,9 @@
+import { User } from '@prisma/client';
+
+export class UserDto {
```
rad20c, line 3: rename it to `UserResponse`

src/user/user.repository.ts
```diff
@@ -0,0 +1,49 @@
+import { Injectable } from '@nestjs/common';
+import { PrismaService } from '../prisma/prisma.service';
+import { Prisma, User } from '@prisma/client';
+import { UserDto } from './dto/user-response';
+
+@Injectable()
+export class UserRepository {
+  constructor(private prisma: PrismaService) {}
+
+  async getUser(where: Prisma.UserWhereUniqueInput): Promise<UserDto> {
```
rad20c, line 10: As discussed, I'd remove `where: Prisma.UserWhereUniqueInput` and split this method into two:
- `getUserById(userId: string)`
- `getUserByEmail(email: string)` (if it's used)

I'd do the same for others.

The reason is to not create a query in service that is calling the repository

src/user/user.repository.ts
```diff
@@ -0,0 +1,49 @@
...
+    return UserDto.from(user);
+  }
+
+  getUserWithPassword(where: Prisma.UserWhereUniqueInput): Promise<User> {
+    return this.prisma.user.findUnique({
+      where,
+    });
+  }
+
+  async createUser(data: Prisma.UserCreateInput): Promise<UserDto> {
```
rad20c, line 24: Here you are returning `Promise<User>` and here `Promise<UserDto>` - I think I'd just return prisma `User` type here. It comes with some disadvantages though but I think it will be better right now.

Use `UserDto` (`UserResponse` after applying suggestion above) in controller when returning data.

src/user/user.service.ts
```diff
@@ -1,68 +1,24 @@
...
-  constructor(private prisma: PrismaService) {}
+  constructor(private readonly userRepository: UserRepository) {}
 
-  async findOneUser(email: string): Promise<User | undefined> {
-    return this.prisma.user.findUnique({
-      where: {
-        email,
-      },
-    });
+  createUser(data: Prisma.UserCreateInput): Promise<UserDto> {
```
rad20c, line 10: same as above

src/auth/auth.service.spec.ts
```diff
@@ -93,14 +93,16 @@ describe('AuthService', () => {
...
-          password: hashed_password,
-          role: Role.USER,
+      jest
+        .spyOn(userRepository, 'getUserWithPassword')
+        .mockImplementation(() => {
+          return Promise.resolve({
+            id: faker.number.int(),
+            email: faker.internet.email(),
+            password: hashed_password,
+            role: Role.USER,
```
rad20c, line 103: I'd use Pascal Case for enums if possible, so `Role.User`.

Also, here you are using enum but above you are passing normal string.

rad20c (changes requested):
check out the comments 👍🏻

src/user/user.repository.ts
rad20c: Before changing the implementation check out [this comment](https://github.com/KubaBaniak/RecipeAppNestjs/pull/25/files#r1265328378).

src/user/user.repository.ts
KubaBaniak: This should be `userId: number` right?

src/user/dto/user-response.ts
KubaBaniak: I'll stick with the first one 😉

src/recipe/recipe.repository.ts
```diff
@@ -0,0 +1,42 @@
...
+@Injectable()
+export class RecipeRepository {
+  constructor(private prisma: PrismaService) {}
+
+  async createRecipe(data: CreateRecipeRequest): Promise<Recipe> {
+    const recipe = await this.prisma.recipe.create({
+      data,
+    });
+
+    return recipe;
```
rad20c, line 15: you don't have to assign result to `recipe` and return it - you can return straight away:
```suggestion
    return this.prisma.recipe.create({
      data,
    });
```

rad20c: implementing above you can also remove `async` (`Promise` in `Promise<Recipe>` stays)

src/user/dto/update-user-request.ts
```diff
@@ -0,0 +1,12 @@
...
+export interface UpdateUserRequest {
+  where: {
+    id: number;
+  };
+  data: {
+    email?: string;
+    password?: string;
+    role?: Role;
+  };
+}
```
rad20c, line 12: shouldn't this be a class with `class-validator` validation applied?

src/user/dto/user-response.ts
```diff
@@ -0,0 +1,18 @@
...
+    public email: string,
+    public role: string,
+    public password?: string,
+  ) {}
+
+  public static from(user: User): UserDto {
+    return new UserDto(user.id, user.email, user.role);
+  }
+
+  public static fromWithPassword(user: User): UserDto {
```
rad20c, line 15: hmm, I'm wondering if it should't be named `withPasswordFrom` because the line in the code will look like this: `UserDto.withPasswordFrom(user);` - so you would read it as `object of type UserDto with password field created from user object`. It could also be read as if we are getting only password from `user` 😅 But I think it's still better than the current name

rad20c (changes requested):
check new comments

src/user/dto/update-user-request.ts
KubaBaniak: Right, to be consistent it should not have such name, but `UpdateUserRequestDto` is kind of counter intuitive for me because I update nothing - I just wanna map fields and return something else. What about `MappedUserRequest` or `UserPayloadRequest`

src/auth/auth.service.ts
```diff
@@ -5,24 +5,28 @@ import {
...
   signIn(signInRequest: SignInRequest): Promise<string> {
     return this.jwtService.signAsync(signInRequest);
   }
 
-  async signUp(signUpRequest: SignUpRequest): Promise<User> {
-    const user = await this.userService.findOneUser(signUpRequest.email);
+  async signUp(signUpRequest: SignUpRequest): Promise<UserPayloadRequest> {
+    const user = await this.userRepository.getUserByEmailWithPassword(
+      signUpRequest.email,
+    );
```
rad20c, line 29: here I think we should return `User` as before

src/auth/dto/sign-up-response.ts
```diff
@@ -1,9 +1,9 @@
-import { User } from '@prisma/client';
+import { UserPayloadRequest } from '../../user/dto/user-request';
 
 export class SignUpResponse {
   constructor(public id: number, public email: string, public role: string) {}
 
-  public static from(user: User): SignUpResponse {
+  public static from(user: UserPayloadRequest): SignUpResponse {
```
rad20c, line 6: check comment above

src/auth/auth.service.ts
KubaBaniak: Actually, when I look at this I'm more confused then before. This at the end calls userService.createUser() which calls userRepository which calls prisma so then Prisma returns User, next userRepository maps it and return same thing, as userService, which passes this thing back to authService (which is UserPayloadRequest). And that's right to me because auth.service should return User without password so controller can send it out.
So should we really change it?

rad20c: ok, so here ideally we should return an object that has only the fields that we need. I don't want to return `UserPayloadRequest` here, because as name suggest it's not a response, so it would be confusing.

PR #26, opened 2023-07-14
feat: generate changelog automatically

KubaBaniak:
### Acceptance criteria:
- When / after deploying commits to the main  branch:
  - CHANGELOG.md  file should be updated or generated
  - version in package.json  and package-lock.json  should be bumped depending on the changes
  - commit should receive a tag with matching version
  - all of the above should be done automatically

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,7 +79,15 @@ jobs:
             E2E_TEST_FILES=$(circleci tests glob "./test/*{e2e-spec}.ts" | circleci tests split --split-by=timings)
             npm test -- $E2E_TEST_FILES --config ./test/jest-e2e.json
       - store_test_results:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
rad20c, line 1: I must have missed that in PR related to CI config - you should install dependencies using `npm ci` command instead of `npm install` / `npm i` command (`npm ci` installs packages from `package-lock.json` while `npm install` / `npm i` installs packages from `package.json` and updates `package-lock.json` if necessary (e.g. if version changed or if any of dependencies' peer dependencies got updated), potentially resulting in a slightly bit different content in `node_modules`

rad20c (approved):
Approved ✅ Hopefully it works 🙂

PR #28, opened 2023-07-24
feat: upload recipes images

KubaBaniak:
### Acceptance criteria:
- Backend exposes an endpoint that can be used to upload any image from the clients' device to AWS S3 bucket
- After image is uploaded to S3, we should store the url in the recipe object
- We should allow uploading image files only


example.env
```diff
@@ -7,3 +7,7 @@ JWT_SECRET=
 BCRYPT_SALT_ROUNDS=
 REDIS_PASSWORD=
 REDIS_URL=
+AWS_S3_BUCKET=
+AWS_S3_ACCESS_KEY=
+AWS_S3_KEY_SECRET=
```
rad20c, line 12: for these 2 environment variables I would use default names used by AWS SDK, that is `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`: https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/loading-node-credentials-environment.html

prisma/schema.prisma
```diff
@@ -28,6 +28,7 @@ model Recipe {
   isPublic    Boolean  @default(false)
   author      User     @relation(fields: [authorId], references: [id], onDelete: Cascade, onUpdate: Cascade)
   authorId    Int
+  imagesUrls  String[] @unique
```
rad20c, line 31: hmm, why `@unique`?

rad20c: also I think it should be optional so that we do not require user to upload image recipe

src/recipe/recipe.controller.ts
```diff
@@ -141,4 +148,38 @@ export class RecipeController {
...
+  })
+  @Post('upload/:id')
+  @UseInterceptors(FilesInterceptor('file'))
+  async uploadFile(
+    @UserId() userId: number,
+    @Param('id', ParseIntPipe) recipeId: number,
+    @UploadedFiles(
+      new ParseFilePipeBuilder()
+        .addMaxSizeValidator({
+          maxSize: 1024 * 1024 * 4,
```
rad20c, line 170: this could be exported to const with a name that tells us what unit it is, e.g. `const MAX_IMAGE_SIZE_IN_MB = 1024 * 1024 * 4`

src/recipe/recipe.controller.ts
```diff
@@ -141,4 +148,38 @@ export class RecipeController {
...
+        })
+        .addFileTypeValidator({
+          fileType: '.(png|jpeg|jpg)',
```
rad20c, line 173: this could also be exported to const (though for now if it's used only here then it should be fine)

src/recipe/s3-bucket.service.ts
```diff
@@ -0,0 +1,50 @@
...
+    return `users/userId:${userId}/recipes/recipeId:${recipeId}/${uuidv4()}`;
+  }
+
+  uploadFile(
+    file: Express.Multer.File,
+    userId: number,
+    recipeId: number,
+  ): string {
+    const key = this.prepareKey(userId, recipeId);
+    this.sendToS3(this.AWS_S3_BUCKET, file.buffer, file.mimetype, key);
```
rad20c, line 26: hmm, I think I would `await` this call

src/recipe/s3-bucket.service.ts
```diff
@@ -0,0 +1,50 @@
...
+    const params = {
+      Bucket: bucket,
+      Key: key,
+      Body: file,
+      ACL: 'public-read',
+      ContentType: mimetype,
+    };
+    const command = new PutObjectCommand(params);
+    try {
+      await this.client.send(command);
```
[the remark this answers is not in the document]
rad20c: note - because you are catching error here you have to `return await` (do not remove `await` here)

rad20c (approved):
looks good 👍🏻

prisma/schema.prisma
KubaBaniak: Unique because url has file name which was generated by `uuidv4`

Prisma automatically initializes `imagesUrls` so there is no need to pass it for e.g. in create request.

PR #29, opened 2023-07-25
feat: return presigned urls instead of normal ones

KubaBaniak:


src/recipe/s3-bucket.service.ts
```diff
@@ -51,4 +55,16 @@ export class S3Service {
       throw new ForbiddenException('Could not send image(s) to storage.');
     }
   }
+
+  getPresignedUrlWithClient(key: string) {
+    const EXPIRE_TIME_IN_SECONDS = 15 * 60;
```
PrzerostNadSens, line 60: set `EXPIRE_TIME_IN_SECONDS` as env

src/recipe/recipe.repository.ts
```diff
@@ -41,13 +41,13 @@ export class RecipeRepository {
     });
   }
 
-  addImageUrls(recipeId: number, urls: string[]): Promise<Recipe> {
+  addImageUrls(recipeId: number, keys: string[]): Promise<Recipe> {
```
PrzerostNadSens, line 44:
```suggestion
  addImageKeys(recipeId: number, keys: string[]): Promise<Recipe> {
```

src/recipe/recipe.repository.ts
```diff
@@ -41,13 +41,13 @@ export class RecipeRepository {
...
     return this.prisma.recipe.update({
       where: {
         id: recipeId,
       },
       data: {
-        imagesUrls: urls,
+        imagesUrls: keys,
```
PrzerostNadSens, line 50: change it to `imageKeys` in DB

PrzerostNadSens (changes requested):

src/recipe/recipe.service.ts
```diff
@@ -134,11 +134,13 @@ export class RecipeService {
       throw new ForbiddenException();
     }
 
-    const urls = await Promise.all(
-      files.map((file) => this.s3Service.uploadFile(file, userId, recipeId)),
+    const keys = await Promise.all(
+      files.map((key) => this.s3Service.uploadFile(key, userId, recipeId)),
```
rad20c, line 138: hmm, wasn't the previous naming better? We have an array of files but the item is called a `key`?

src/recipe/s3-bucket.service.ts
```diff
@@ -51,4 +55,15 @@ export class S3Service {
       throw new ForbiddenException('Could not send image(s) to storage.');
     }
   }
+
+  getPresignedUrlWithClient(key: string) {
```
rad20c, line 59: what `Client`? I think this part is not necessary in method name. It can be either `getPresignedUrl` or `getPresignedUrlForKey`

rad20c (approved):

src/recipe/s3-bucket.service.ts
KubaBaniak: In aws docs there were 2 functions - one using client, and one not. Therefore because I'm using client here, I picked this name here. But in this case you're right. I don't need to specify that here.

PR #30, opened 2023-07-25
chore: update prisma to v5.0.0

KubaBaniak:
### Acceptance criteria:
- Prisma is updated to the newest version.
- Every database related method works as expected.


src/main.ts
```diff
@@ -13,6 +13,8 @@ async function bootstrap() {
     }),
   );
 
+  app.enableShutdownHooks();
```
rad20c, line 16: hmm, I'm wondering if we currently need it - does the application crash without it on shutdown?

rad20c (approved):

src/main.ts
rad20c: if we don't need it then we should remove it. Especially as it consumes resources: https://docs.nestjs.com/fundamentals/lifecycle-events

PR #32, opened 2023-07-27
task: deploy to web

KubaBaniak:


.circleci/config.yml
```diff
@@ -23,27 +23,32 @@ commands:
...
+            PGADMIN_EMAIL="$PGADMIN_EMAIL"
+            PGADMIN_PASSWORD="$PGADMIN_PASSWORD"
+            JWT_SECRET="$JWT_SECRET"
+            BCRYPT_SALT="$BCRYPT_SALT"
+            REDIS_PASSWORD="$REDIS_PASSWORD"
+            REDIS_URL="$REDIS_URL"
+            EOF
+            cat .env > test.env
+            cat .env
+            cat test.env
```
rad20c, line 51: not sure if you want to print the contents at the end (values will be masked but still), unless you want to make sure the file has been written correctly. I would remove this though.

.circleci/config.yml
```diff
@@ -23,27 +23,32 @@ commands:
...
+            POSTGRES_USER="$POSTGRES_USER"
+            POSTGRES_PASSWORD="$POSTGRES_PASSWORD"
+            PGADMIN_EMAIL="$PGADMIN_EMAIL"
+            PGADMIN_PASSWORD="$PGADMIN_PASSWORD"
+            JWT_SECRET="$JWT_SECRET"
+            BCRYPT_SALT="$BCRYPT_SALT"
+            REDIS_PASSWORD="$REDIS_PASSWORD"
+            REDIS_URL="$REDIS_URL"
+            EOF
+            cat .env > test.env
```
rad20c, line 49: Not sure why you need to create both `.env` and `test.env`? I thought we only need to create `test.env` for `jest` tests, and for normal builds we could fetch envs from CI? (might be mistaken though)

.circleci/config.yml
```diff
@@ -23,27 +23,32 @@ commands:
...
       - run:
           name: Run images of services declared in docker-compose.yml
-          command: docker-compose up -d
-  create_test_env:
+          command: docker-compose --profile dependencies up -d
+  create_env:
     steps:
       - run:
-          name: Create test.env file
+          name: Create .env file
```
rad20c, line 36: If you have to create 2 environment files then modify names here accordingly, because right now you are not creating just `.env` file only.

.circleci/config.yml
```diff
@@ -109,17 +114,47 @@ jobs:
...
+          name: Deploy over SSH
+          command: |
+            ssh -o StrictHostKeyChecking=accept-new $SSH_USER@$SSH_HOST "source .env && ./recipe_app_setup.sh"
 
 workflows:
   build_and_test:
     jobs:
-      - build
+      - build:
+          context: test
```
rad20c, line 151: If you need this context, name it in a descriptive way so that we know what it's responsible for

.circleci/config.yml
```diff
@@ -109,17 +114,47 @@ jobs:
...
-      - run: echo "!TODO"
+      - add_ssh_keys:
+          fingerprints:
+            - '97:69:fe:70:ca:84:59:06:98:ab:9c:ec:d5:26:49:8c'
+      - checkout
+      - setup_remote_docker:
+          version: 20.10.14
+      - run:
+          name: Build App image
+          command: docker build -t kuba1212/recipeapp-docker .
```
rad20c, line 125: `kuba1212/recipeapp-docker` - I would move it to CircleCI environment variable

.circleci/config.yml
```diff
@@ -109,17 +114,47 @@ jobs:
...
+          command: docker save kuba1212/recipeapp-docker | gzip > appimage.tar.gz
+      - create_env
+      - run:
+          name: Send production env
+          command: |
+            scp -o StrictHostKeyChecking=accept-new ./.env $SSH_USER@$SSH_HOST:.env
+      - run:
+          name: Send image to server
+          command: |
+            scp -o StrictHostKeyChecking=accept-new ./appimage.tar.gz $SSH_USER@$SSH_HOST:appimage.tar.gz
```
rad20c, line 137: if we send the built image to a docker hub - do we need to upload this image to the server? Shouldn't the server be able to download the image from the docker hub?

.circleci/config.yml
```diff
@@ -130,6 +165,8 @@ workflows:
   deploy:
     jobs:
       - deploy:
+          context: production
```
rad20c, line 168: as above - a better name would be appreciated

.circleci/config.yml
rad20c: Take a note that context is shared across all CI projects on the account. So it should at minimum include a project name.

.circleci/config.yml
KubaBaniak: That's because later, when I run `setup_docker_compose_dependencies` to setup db, docker look sat .env file I think.

package.json
```diff
@@ -38,7 +39,7 @@
     "@nestjs/platform-express": "10.0.5",
     "@nestjs/swagger": "7.1.1",
     "@prisma/client": "5.0.0",
-    "bcrypt": "5.1.0",
+    "bcryptjs": "2.4.3",
```
rad20c, line 42: I remember you were mentioning an issue with `bcrypt` - could you elaborate here what was the issue and why did we change it to `bcryptjs`. If I remember correctly it was connected to using the `alpine` linux distribution

nest-cli.json
```diff
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/nest-cli",
   "collection": "@nestjs/schematics",
-  "sourceRoot": "src",
+  "sourceRoot": "/",
```
rad20c, line 4: out of curiosity - what was happening that you had to make this change?

KubaBaniak: It was an attempt to fix the `npm run build` problem we have talked about. Forgot to undo this

package.json
KubaBaniak: On `npx run build` bcrypt_module could not be generated. People suggested using `bcryptjs` or installing `gyp` (https://github.com/kelektiv/node.bcrypt.js/issues/800). Also, I don't remember where but people pointed out that it is due to the fact `bcrypt` uses C++ (https://codeforgeek.com/bcrypt-vs-bcryptjs/) where bcryptjs uses plan JS.

.circleci/config.yml
KubaBaniak: This is `console.log()` like accident 😅

.circleci/config.yml
KubaBaniak: Talked about this on Backend meeting

docker-compose.yml
```diff
@@ -32,9 +34,22 @@ services:
     command: redis-server --save 20 1 --loglevel warning --requirepass '${REDIS_PASSWORD}'
     volumes:
       - cache:/data
+    profiles:
+      - dependencies
+    networks:
+      - recipe-app-network
+  application:
+    image: kuba1212/recipeapp-docker
+    env_file: .env
```
rad20c, line 43: kudos to Arek for this:

```suggestion
  application:
    image: kuba1212/recipeapp-docker
    build: .
    env_file: .env
```

it will work like this:
1. if we have a `kuba1212/recipeapp-docker` image available then it will use this image
2. if not it will try to build it using Dockerfile

This way we can also use this docker-compose file locally for running the newest backend app in the docker too.

If we want to make sure that on the server we always fetch the image (because we don't have source files or Dockerfile) we should add `--no-build` flag, e.g. `docker-compose up --no-build`.

On the other hand, locally you can force to build docker image instead of pulling it by using `--build <service_name>`, e.g. `docker compose up --build <service_name>` (you should test / verify it though)

.circleci/config.yml
rad20c: ok, if that's the case then fine 👍🏻 Would be good to be sure of that though 🙂

rad20c (approved):

PR #36, opened 2023-08-03
task: add user in dockerfile

KubaBaniak:


Dockerfile
```diff
@@ -1,29 +1,33 @@
...
+COPY --chown=node:node package*.json ./
+COPY --chown=node:node tsconfig*.json ./
+COPY --chown=node:node src ./src
+COPY --chown=node:node prisma ./prisma/
 
 RUN npm ci
 
 RUN npm run build
 
 FROM node:18-alpine as production
```
rad20c, line 16: here you have `as production` and above `as development` which is kind of misleading. I would name the above `as builder` and here `as app`

rad20c (approved):

PR #37, opened 2023-08-03
fix: fix fetching public and private recipes

KubaBaniak:
### **Issue** 
Endpoint would fetch only public recipes, while it should fetch all public recipes as well as all private recipes of user who requested these recipes.

### DIFF
--- a/src/recipe/recipe.repository.ts
+++ b/src/recipe/recipe.repository.ts
@@ -56,10 +56,10 @@ export class RecipeRepository {
     return this.prisma.recipe.findMany();
   }
 
-  getAllRecipesForBasicUser(authorId: number): Promise<Recipe[]> {
+  getAllAvailableRecipesForUser(authorId: number): Promise<Recipe[]> {
     return this.prisma.recipe.findMany({
       where: {
-        OR: [{ isPublic: true }, { authorId, isPublic: false }],
+        isPublic: true,
       },
     });
   }
--- a/src/recipe/recipe.repository.ts
+++ b/src/recipe/recipe.repository.ts
@@ -56,7 +56,7 @@ export class RecipeRepository {
     return this.prisma.recipe.findMany();
   }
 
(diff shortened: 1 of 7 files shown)


src/recipe/recipe.repository.ts
```diff
@@ -56,10 +56,10 @@ export class RecipeRepository {
     return this.prisma.recipe.findMany();
   }
 
-  getAllPublicRecipes(): Promise<Recipe[]> {
+  getAllPublicRecipes(authorId: number): Promise<Recipe[]> {
```
rad20c, line 59: Naming is kind of misleading because here we are getting both public recipes and private recipes but belonging to the `authorId`. It should be named better - I don't have a good idea right now but perhaps something along `getAllAvailableRecipesForUser(userId)` or `getAllRecipesTheUserHasAccessTo(userId)` or `getAllPublicAndOwnerRecipes(ownerId)` or something else.

src/recipe/recipe.service.ts
```diff
@@ -77,10 +77,10 @@ export class RecipeService {
 
   async fetchAllRecipes(userId: number): Promise<Recipe[]> {
     const user = await this.userRepository.getUserById(userId);
-    if (user.role === Role.ADMIN) {
+    if (user?.role === Role.ADMIN) {
```
rad20c, line 80: why user might not have `role` defined?

rad20c (changes requested):
Check comments + if the task doesn't have any JIRA / ClickUp ticket associated, please write a short description in the task what was the issue and how is this PR fixing it.

src/recipe/recipe.service.ts
KubaBaniak: Thats interesting because I accidentally made a request with a token from a user who has been deleted (I created user > sign in > `prisma migrate reset` > fetch with the old token). This caused 500 error which was caused by the fact, that `undefined` doesn't have `role` field.

rad20c (approved):

PR #44, opened 2023-08-09
fix: send presigned urls on fetch instead of keys

KubaBaniak:
### Issue: 
GET /recipes and GET /recipes/:recipeId return imageKeys instead of links

### DIFF
--- a/src/recipe/recipe.service.ts
+++ b/src/recipe/recipe.service.ts
@@ -20,6 +20,21 @@ export class RecipeService {
     private readonly s3Service: S3Service,
   ) {}
 
+  getImageUrlsToUrls(recipe: Recipe): Promise<string[]> {
+    return Promise.all(
+      recipe.imageKeys.map((key) => this.s3Service.getPresignedUrl(key)),
+    );
+  }
+
+  getImageUrlsOfManyRecipes(recipes: Recipe[]): Promise<Recipe[]> {
+    return Promise.all(
+      recipes.map(async (recipe) => {
+        recipe.imageKeys = await this.getImageUrlsToUrls(recipe);
+        return recipe;
+      }),
+    );
+  }
+
   async createRecipe(
     userId: number,
     data: CreateRecipeRequest,
@@ -55,6 +70,8 @@ export class RecipeService {
       throw new ForbiddenException();
     }
 
+    recipe.imageKeys = await this.getImageUrlsToUrls(recipe);
+
     return recipe;
   }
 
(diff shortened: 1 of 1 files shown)


src/recipe/recipe.service.ts
```diff
@@ -20,6 +20,21 @@ export class RecipeService {
     private readonly s3Service: S3Service,
   ) {}
 
+  getImageUrlsToUrls(recipe: Recipe): Promise<string[]> {
```
rad20c, line 23: "image urls to urls"? Sounds strange to me. Should it be perhaps image keys to urls?

rad20c: I would rename it to `getPresignedUrlsforRecipeImages` or something like that

src/recipe/recipe.service.ts
```diff
@@ -20,6 +20,21 @@ export class RecipeService {
...
+    );
+  }
+
+  getImageUrlsOfManyRecipes(recipes: Recipe[]): Promise<Recipe[]> {
+    return Promise.all(
+      recipes.map(async (recipe) => {
+        recipe.imageKeys = await this.getImageUrlsToUrls(recipe);
+        return recipe;
+      }),
+    );
```
rad20c, line 35: check if you can use spread operator here (I'm not 100% sure):

```suggestion
    return Promise.all(
      recipes.map(async (recipe) => ({ ...recipe, imageKeys: await this.getImageUrlsToUrls(recipe) })),
    );
```

src/recipe/recipe.service.ts
```diff
@@ -69,18 +86,28 @@ export class RecipeService {
     const principal = await this.userRepository.getUserById(principalId);
 
     if (author.id === principal.id || principal.role == Role.ADMIN) {
-      return this.recipeRepository.getUsersRecipes(authorId);
+      const recipes = await this.recipeRepository.getUsersRecipes(authorId);
+      const recipesWithUrls = this.getImageUrlsOfManyRecipes(recipes);
+      return recipesWithUrls;
```
rad20c, line 91: you can return this immediately without the need for creating a variable

src/recipe/recipe.service.ts
```diff
@@ -69,18 +86,28 @@ export class RecipeService {
...
+    return recipesWithUrls;
   }
 
   async fetchAllRecipes(userId: number): Promise<Recipe[]> {
     const user = await this.userRepository.getUserById(userId);
     if (user.role === Role.ADMIN) {
-      return this.recipeRepository.getAllRecipes();
+      const recipes = await this.recipeRepository.getAllRecipes();
+      const recipesWithUrls = this.getImageUrlsOfManyRecipes(recipes);
+      return recipesWithUrls;
```
rad20c, line 104: as above

src/recipe/recipe.service.ts
```diff
@@ -69,18 +86,28 @@ export class RecipeService {
...
     } else {
-      return this.recipeRepository.getAllAvailableRecipesForUser(userId);
+      const recipes = await this.recipeRepository.getAllAvailableRecipesForUser(
+        userId,
+      );
+      const recipesWithUrls = this.getImageUrlsOfManyRecipes(recipes);
+      return recipesWithUrls;
```
rad20c, line 110: as above

src/recipe/recipe.service.ts
```diff
@@ -55,6 +70,8 @@ export class RecipeService {
       throw new ForbiddenException();
     }
 
+    recipe.imageKeys = await this.getImageUrlsToUrls(recipe);
+
     return recipe;
```
rad20c, line 75: again - check if you can use spread operator here. If `recipe` is an interface then you should be able to, if class object then I don't think so. Also the await might be a problem but I think it should pass.

```suggestion
    return { ...recipe, imageKeys:  await this.getImageUrlsToUrls(recipe) }
```

rad20c (approved):

PR #45, opened 2023-08-10
task: auth sockets

KubaBaniak:


src/auth/auth.service.ts
```diff
@@ -23,6 +23,10 @@ export class AuthService {
     private readonly jwtService: JwtService,
   ) {}
 
+  async verifyJwt(jwtToken: string): Promise<any> {
```
PrzerostNadSens, line 26: change this type to `unknown` or to a specific type if you can determine it, try not to use it `any`

PrzerostNadSens: you can add interface `AccessTokenPayload`

src/websocket/notification.gateway.ts
```diff
@@ -1,35 +1,60 @@
...
-} from '@nestjs/websockets';
+import { Injectable, UnauthorizedException } from '@nestjs/common';
+import { WebSocketGateway, WebSocketServer } from '@nestjs/websockets';
 import { Server, Socket } from 'socket.io';
+import { AuthService } from '../auth/auth.service';
+import { UserRepository } from '../user/user.repository';
 
 @Injectable()
 @WebSocketGateway()
 export class NotificationGateway {
```
PrzerostNadSens, line 9: in my opinion, it is better to change the name of it to `WebSocketGateway` (and rename the entire file to `WebSocketGateway`)

```suggestion
export class WebSocketGateway {
```

PrzerostNadSens: you should not connect WebSocket with only notifications 

it is better call it as  WebSocket events

src/websocket/notification.gateway.ts
```diff
@@ -1,35 +1,60 @@
...
 
-  getNotificationRoomName(): string {
-    return `recipe-notification-room`;
+  getRoomWithId(userId: number): string {
+    return `notification-userId-${userId}`;
+  }
+
+  async handleConnection(socket: Socket) {
+    const token: string =
+      socket.handshake.headers['authorization'].split(' ')[1];
```
PrzerostNadSens, line 24:
```suggestion
      socket.handshake.headers.authorization;
```
it should also work, you can try it

src/websocket/notification.gateway.ts
```diff
@@ -1,35 +1,60 @@
...
-  @SubscribeMessage('room')
-  joinRoom(@ConnectedSocket() client: Socket): void {
-    client.join(this.getNotificationRoomName());
+  addUserToNotificationRoom(socket: Socket, userId: number): void {
+    socket.join(this.getRoomWithId(userId));
   }
 
-  sendRecipeNotification(message: string): boolean {
-    return this.server.emit('notification', message);
+  sendRecipeNotification(message: string, authorId: number): boolean {
```
PrzerostNadSens, line 48:
```suggestion
  sendRecipeWebsocketEvent(message: string, authorId: number): boolean {
```

src/websocket/notification.gateway.ts
```diff
@@ -1,35 +1,60 @@
...
+
+  private dissconnectOnAuth(socket: Socket) {
+    socket.emit('Error', new UnauthorizedException());
+    socket.disconnect();
   }
 
-  @SubscribeMessage('room')
-  joinRoom(@ConnectedSocket() client: Socket): void {
-    client.join(this.getNotificationRoomName());
+  addUserToNotificationRoom(socket: Socket, userId: number): void {
```
PrzerostNadSens, line 44:
```suggestion
  addUserToWebsocketEventRoom(socket: Socket, userId: number): void {
```

PrzerostNadSens (approved):

src/websocket/notification.gateway.ts
KubaBaniak: Maybe something like `WebSocketEventGateway()` is good enough?

PrzerostNadSens: I think`WebSocketEventGateway()` will fit

PR #46, opened 2023-08-10
task: public api with pat

KubaBaniak:
### Acceptance criteria:
- User should be able to generate his personal access token (PAT)
- User should be able to use his PAT instead of Bearer token to access endpoints for:
   - fetching list or recipes
   - fetching a specific recipe
   - updating a recipe
   - deleting a recipe
- Generated PAT should not expire


src/auth/auth.service.ts
```diff
@@ -36,7 +36,28 @@ export class AuthService {
...
+    return await this.jwtService.signAsync(
+      {
+        id: user.id,
+        email: user.email,
+      },
+      { expiresIn: `${process.env.JWT_EXPIRY_TIME}s` },
+    );
+  }
+
+  async createPAT(userId: number): Promise<string> {
```
rad20c, line 48: I'd prefer using full name -> `Personal Access Token` here

src/auth/auth.service.ts
```diff
@@ -36,7 +36,28 @@ export class AuthService {
...
+    const validPatToken = await this.userRepository.getUserValidPatWithId(
```
rad20c, line 49:
```suggestion
    const validPatToken = await this.userRepository.getValidPatForUserId(
```

rad20c: `PatToken` sounds strange -> "Personal Access Token Token"

src/auth/dto/create-pat-response.ts
```diff
@@ -0,0 +1,11 @@
+export class CreatePatResponse {
+  public patToken: string;
+
+  constructor(patToken: string) {
```
rad20c, line 4: patToken again - I'd just name it personalAccessToken

prisma/migrations/20230811101255_add_pta/migration.sql
```diff
@@ -0,0 +1,14 @@
+-- CreateTable
+CREATE TABLE "PTA" (
```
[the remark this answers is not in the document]
rad20c: I would name the table `PersonalAccessTokens`

example.env
```diff
@@ -5,6 +5,7 @@ PGADMIN_EMAIL=
 PGADMIN_PASSWORD=
 JWT_EXPIRY_TIME=
 JWT_SECRET=
+JWT_EXPIRY_TIME=
```
rad20c, line 8: duplicated entry `JWT_EXPIRY_TIME`

docs/CHANGELOG.md
rad20c, line 1: hmm, this file should not be in this PR (there should be no changes to this file in the PR)

prisma/migrations/20230811101958_rename_to_pat/migration.sql
```diff
@@ -0,0 +1,26 @@
...
+
+*/
+-- DropForeignKey
+ALTER TABLE "PTA" DROP CONSTRAINT "PTA_userId_fkey";
+
+-- DropTable
+DROP TABLE "PTA";
+
+-- CreateTable
+CREATE TABLE "PAT" (
```
rad20c, line 14: ok, now I see - but again I would name it `PersonalAccessTokens` - when you dive in into the code for the first time you might not know if PAT is PersonalAccessToken or Port Address Translation or Portland Association of Teachers

prisma/schema.prisma
```diff
@@ -28,7 +29,15 @@ model Recipe {
...
-  imageKeys  String[] @unique
+  imageKeys   String[] @unique
+}
+
+model PAT {
+  id            Int       @id @default(autoincrement())
+  user          User      @relation(fields: [userId], references: [id])
+  userId        Int
+  token         String
+  invalidatedAt DateTime?
```
rad20c, line 40: I'd also add `createdAt` field

src/auth/auth.module.ts
```diff
@@ -8,6 +8,7 @@ import { LocalStrategy } from './strategies/local.strategy';
 import { JwtStrategy } from './strategies/jwt.strategy';
 import { PassportModule } from '@nestjs/passport';
 import { LocalAuthGuard } from './guards/local-auth.guard';
+import { PtaJwtStrategy } from './strategies/pta-jwt.strategy';
```
rad20c, line 11: pta again

src/auth/auth.service.ts
```diff
@@ -36,7 +36,28 @@ export class AuthService {
       throw new NotFoundException();
     }
 
-    return await this.jwtService.signAsync({ id: user.id, email: user.email });
+    return await this.jwtService.signAsync(
```
rad20c, line 39: no need to await on return

src/auth/auth.service.ts
```diff
@@ -36,7 +36,28 @@ export class AuthService {
...
+    );
+    if (validPatToken) {
+      this.userRepository.invalidateUserPat(userId);
+    }
+    const token = await this.jwtService.signAsync({
+      id: userId,
+      type: 'PAT',
+    });
+    const patToken = await this.userRepository.createPat(userId, token);
+    return patToken.token;
```
rad20c, line 60:
```suggestion
    const { token } = await this.userRepository.createPat(userId, token);
    return token;
```

I think it's more clear. `patToken.token` sounded strange

rad20c: I'd rename `createPat` to `createPersonalAccessToken` or `generatePersonalAccessToken`

src/auth/guards/pta-auth-api.guard.ts
```diff
@@ -0,0 +1,5 @@
+import { Injectable } from '@nestjs/common';
+import { AuthGuard } from '@nestjs/passport';
+
+@Injectable()
+export class ApiAuthGuard extends AuthGuard(['jwt.pta', 'jwt.bearer']) {}
```
rad20c, line 5: pta?

src/auth/strategies/pta-jwt.strategy.ts
```diff
@@ -0,0 +1,18 @@
+import { ExtractJwt, Strategy } from 'passport-jwt';
+import { PassportStrategy } from '@nestjs/passport';
+import { Injectable } from '@nestjs/common';
+import { jwtConstants } from '../constants';
+import { PtaJwtPayload } from './payloads/pta-jwt-token.payload';
+
+@Injectable()
+export class PtaJwtStrategy extends PassportStrategy(Strategy, 'jwt.pta') {
```
rad20c, line 8: I'd name it `PersonalAccessTokenStrategy`

src/auth/strategies/jwt.strategy.ts
```diff
@@ -5,11 +5,10 @@ import { jwtConstants } from '../constants';
 import { JwtPayload } from './payloads/jwt-token.payload';
 
 @Injectable()
-export class JwtStrategy extends PassportStrategy(Strategy) {
+export class JwtStrategy extends PassportStrategy(Strategy, 'jwt.bearer') {
```
rad20c, line 8: I think we can rename it to e.g. `BearerStrategy`

src/auth/strategies/local.strategy.ts
```diff
@@ -15,7 +15,7 @@ export class LocalStrategy extends PassportStrategy(Strategy) {
     const user = await this.authService.validateUser(data);
 
     if (!user) {
-      throw new UnauthorizedException();
+      throw new NotFoundException();
```
rad20c, line 18: I think the `UnauthorizedException()` is an appropriate error here. Now we are explicitly telling user that "hey, this user does not exist in the system", thus enabling malicious actors to perform a user enumeration attack. Note that `this.authService.validateUser(data);` method can return `UnauthorizedException()` on its own. We should return `UnauthorizedException()` here

src/auth/strategies/pta-jwt.strategy.ts
```diff
@@ -0,0 +1,18 @@
...
+  constructor() {
+    super({
+      jwtFromRequest: ExtractJwt.fromAuthHeaderAsBearerToken(),
+      secretOrKey: jwtConstants.secret,
+    });
+  }
+  validate(payload: PtaJwtPayload): PtaJwtPayload {
```
rad20c, line 15: pta?

src/metadata.ts
```diff
@@ -153,6 +164,9 @@ export default async () => {
             AuthController: {
               signIn: { type: t['./auth/dto/sign-in-response'].SignInResponse },
               signUp: { type: t['./auth/dto/sign-up-response'].SignUpResponse },
+              generatePAT: {
+                type: t['./auth/dto/create-pat-response'].CreatePatResponse,
```
rad20c, line 168: hmm, where did `generatePAT` came from? You have `createPat` / `createPAT` everywhere

src/user/user.repository.ts
```diff
@@ -48,4 +47,40 @@ export class UserRepository {
...
+      data: {
+        token,
+        user: {
+          connect: {
+            id: userId,
+          },
+        },
+      },
+    });
+    return test;
```
rad20c, line 62: you should be able to immediately return here. Also I think `test` field was for test purposes right? 😉

src/user/user.repository.ts
```diff
@@ -48,4 +47,40 @@ export class UserRepository {
...
+        userId,
+        invalidatedAt: {
+          equals: null,
+        },
+      },
+    });
+  }
+
+  async invalidateUserPat(userId: number): Promise<void> {
+    await this.prisma.pAT.updateMany({
```
rad20c, line 77: you can remove `async` and `await` and keep `Promise<void>` - you will still be able to `await` it where you call this method

test/recipe.e2e-spec.ts
```diff
@@ -24,6 +24,7 @@ describe('RecipeController (e2e)', () => {
   let authService: AuthService;
   let user: User;
   let accessToken: string;
+  let patToken: string;
```
rad20c, line 27: "patToken" sound strange - I'd rename to `personalAccessToken`

rad20c (changes requested):
check comments

src/metadata.ts
KubaBaniak: this is the name of controller method

src/user/user.repository.ts
KubaBaniak: I've tested it out and without `async` `await` in `user.repository` it doesn't update `invalidatedAt` field (I awaited this in `auth.service`).

src/auth/strategies/local.strategy.ts
KubaBaniak: Ok I get it. If so then should I also change the Exception within `signIn` method in `auth.service`?

prisma/schema.prisma
```diff
@@ -11,11 +11,12 @@ datasource db {
...
-  email    String   @unique
-  password String
-  role     Role     @default(USER)
-  recipes  Recipe[]
+  id                  Int                   @id @default(autoincrement())
+  email               String                @unique
+  password            String
+  role                Role                  @default(USER)
+  recipes             Recipe[]
+  personalAccessToken PersonalAccessToken[]
```
rad20c, line 19: hmm, I wouldn't add `personalAccessToken` to the `User` table. You already have `userId` in `PersonalAccessToken` table - this should be enough.

src/auth/strategies/local.strategy.ts
rad20c: if the logic is exactly the same as above, then yes, I would change it 👍🏻

src/user/user.repository.ts
rad20c: hmm, have you tried returning it? I mean adding `return` statement.

However, we might not want to return anything (so leaving `Promise<void>` intact) -in this case what you have now is correct 👍🏻

src/metadata.ts
rad20c: ah I see - how about we unify the method name for creating the personal access token everywhere (in controller and service)?

src/auth/auth.service.ts
rad20c: Regarding suggestion - sure, but I'd remove `ToDb` part.

But now I'm thinking - why do we have `createPat` method on `userRepository`? Shouldn't we call `personalAccessTokenRepository` instead?

prisma/schema.prisma
KubaBaniak: I can't remove `personalAccessToken` here.
Prisma throws this error:
```
Error validating field 'user' in model 'PersonalAccessToken': 
The relation field 'user' on model 'PersonalAccessToken' is missing an opposite relation field on the model 'User'`
```

rad20c (approved):

PR #47, opened 2023-08-10
fix: change imageKeys property to imageUrls while fetching

KubaBaniak:
## Issue: 

`imageKeys` property name should be called something else while fetching

### DIFF
--- a/src/recipe/recipe.service.ts
+++ b/src/recipe/recipe.service.ts
@@ -4,7 +4,11 @@ import {
   ForbiddenException,
 } from '@nestjs/common';
 import { Recipe } from '@prisma/client';
-import { CreateRecipeRequest, UpdateRecipeRequest } from './dto';
+import {
+  CreateRecipeRequest,
+  RecipeWithUrls,
+  UpdateRecipeRequest,
+} from './dto';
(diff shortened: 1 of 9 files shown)


src/recipe/recipe.service.ts
```diff
@@ -69,17 +76,18 @@ export class RecipeService {
     if (user.id != recipe.authorId && !recipe.isPublic) {
       throw new ForbiddenException();
     }
+    const { imageKeys, ...recipeWithoutImageKeys } = recipe;
```
PrzerostNadSens, line 79: please change it, you don't need to get imageKeys

PrzerostNadSens (approved):

src/recipe/recipe.service.ts
KubaBaniak: How would You approach this then? Because i want to extract everything but `imageKeys`

PrzerostNadSens: you can use lodash to omit `imageKeys` but if you don't use this library in your project then I don't need to add it for this one case and it's better to leave it as it is

const recipeWithoutImageKeys = lodash.omit(recipe, ['imageKeys']);

PR #48, opened 2023-08-17
feat: add webhooks

KubaBaniak:
### Acceptance criteria:
- User can add his own webhook urls.
- User should be able to define an Authorization  header that should be sent when webhook is triggered.
- For a better user experience - user should be able to set a name for each webhook.
- User should be able to define up to max 5 webhooks.
- Webhooks can be triggered on a following events:
  - when a new recipe is created
  - when a recipe has been updated
  - when a recipe has been deleted
- Webhooks should expect a 200 OK  response.
- If the response has any other response than 200 OK  then it should be considered as not delivered. There should be a retention policy of up to 5 or more retries implemented.


prisma/migrations/20230814090235_add_webhook_as_relation/migration.sql
```diff
@@ -0,0 +1,21 @@
...
+  Warnings:
+
+  - You are about to drop the column `webhooks` on the `User` table. All the data in the column will be lost.
+
+*/
+-- AlterTable
+ALTER TABLE "User" DROP COLUMN "webhooks";
+
+-- CreateTable
+CREATE TABLE "Webhook" (
```
rad20c, line 11: hmm, I've noticed it a bit to late I guess - I think the table names should be in plural - so `Webhooks` instead of `Webhook`, `Users` instead of `User` and `Recipes` instead of `Recipe`

prisma/schema.prisma
```diff
@@ -11,11 +11,12 @@ datasource db {
 }
 
 model User {
-  id       Int      @id @default(autoincrement())
-  email    String   @unique
+  id       Int       @id @default(autoincrement())
+  email    String    @unique
   password String
-  role     Role     @default(USER)
+  role     Role      @default(USER)
```
rad20c, line 17: hmm, were these indents intended?

src/app.module.ts
```diff
@@ -7,6 +7,7 @@ import { RecipeModule } from './recipe/recipe.module';
 import { PrismaModule } from './prisma/prisma.module';
 import { RedisCacheModule } from './cache/redis-cache.module';
 import { WebSocketEventModule } from './websocket/websocket-event.module';
+import { WebhookModule } from './webhook/webhook.module';
```
rad20c, line 10: the folder should also be in plural - "webhooks". I'd leave it for some other PR as there might be a lot of changes in the code.

src/webhook/webhook.service.ts
```diff
@@ -0,0 +1,116 @@
...
+  async recipeDeleted(userId: number, data: Recipe) {
+    const userWebhooks = await this.patRepository.getAllWebhooksByUserId(
+      userId,
+    );
+    userWebhooks.forEach((webhook) => {
+      if (webhook.type === 'DELETE') {
+        this.sendToWebhook(webhook.url, data, webhook.token);
+      }
+    });
+  }
```
rad20c, line 115: hmm, I think it's a bit too much redundancy and we are breaking the DRY principle here (you have 3x almost the same code repeated in the same file). I'd do it differently:

- have one method that sends webhook events - in your case `sendToWebhook`.
- I'd create an enum for events like this:

```typescript
enum WebhookEvent {
  RecipeCreated = "recipe_created",
  RecipeUpdated = "recipe_updated",
  RecipeDeleted = "recipe_deleted"
}
```
- `sendToWebhook` method should accept `WebhookEvent` as parameter
- looking at the implementation we would need an additional method - e.g. `sendWebhookEvent`, where we will include this part: `userWebhooks.forEach((webhook) => { ... })`

rad20c: A word on naming the events:

Please note that this feature is something that our customers integrate with. We have no control over these integrations. Thus we have to be careful when making any changes, as it might break customer's integration. When we really have to make a breaking change, we have to create a `v2` webhook, implement the possibility of letting user decide wich version to use, deprecate `v1` webhook and give users time to migrate to `v2`. All of this takes a lot of time.

Here we should immediately change event names - we have now `CREATE`, `UPDATE` and `DELETE`. What if we want to also have events for when a new user was created? We would create (I persume) a `USER_CREATED` event. But now it would be good to change `CREATE` to `RECIPE_CREATED` as otherwise it might be confusing. But if users are already using our webhooks - we cannot do this easily.

prisma/schema.prisma
KubaBaniak: Yep, they were generated by `npx prisma format`

src/webhook/webhook.service.ts
KubaBaniak: I understand your idea about this implementation, but shouldn't `sendWebhookEvent` accept parameter `WebhookEvent`? I have to specify in `recipe.service` to which type of webhook event should be send, therefore `sendWebhookEvent` should process it and then `sendToWebhook` just sends it.

rad20c: yes, `sendWebhookEvent` should accept `WebhookEvent` as parameter 👍🏻

src/user/user.repository.ts
```diff
@@ -48,4 +49,40 @@ export class UserRepository {
...
+    });
+  }
+
+  async deleteUserWebhookByName(webhookId: number): Promise<void> {
+    await this.prisma.webhook.delete({
+      where: {
+        id: webhookId,
+      },
+    });
+  }
```
rad20c, line 87: These should not be in `userRepository`

src/user/user.repository.ts
```diff
@@ -48,4 +49,40 @@ export class UserRepository {
...
+
+  async getWebhookById(userId: number): Promise<Webhook> {
+    return this.prisma.webhook.findUnique({
+      where: {
+        id: userId,
+      },
+    });
+  }
+
+  async deleteUserWebhookByName(webhookId: number): Promise<void> {
```
rad20c, line 81: the method name says `ByName` but in parameter you have `webhoohId`

src/user/user.repository.ts
```diff
@@ -48,4 +49,40 @@ export class UserRepository {
...
+
+  async getAllWebhooksByUserId(userId: number): Promise<Webhook[]> {
+    return this.prisma.webhook.findMany({
+      where: {
+        userId,
+      },
+    });
+  }
+
+  async getWebhookById(userId: number): Promise<Webhook> {
```
rad20c, line 73: the parameter name is wrong - it should be `webhookId`

src/webhook/dto/webhook-response.ts
```diff
@@ -0,0 +1,3 @@
+import { Webhook } from '@prisma/client';
+
+export type ListWebhooksDto = Omit<Webhook, 'userId'>;
```
rad20c, line 3: hmm, it's kind of risky

advantages:
- you have few lines of code
- you do not have to update this DTO when new fields are added to `Webhook` table

disadvantages:
- you do not have to update this DTO when new fields are added to `Webhook` table - this poses a risk that we leak some data we do not want to. If its used within service then perhaps it's ok.

src/webhook/webhook.controller.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+      controllers: [WebhookController],
+      providers: [WebhookService, WebhookRepository, PrismaService],
+    }).compile();
+
+    controller = module.get<WebhookController>(WebhookController);
+  });
+
+  it('should be defined', () => {
+    expect(controller).toBeDefined();
+  });
```
rad20c, line 23: would be good to add some tests

src/webhook/webhook.controller.ts
```diff
@@ -0,0 +1,51 @@
...
+  async deleteWebhook(
+    @UserId() userId: number,
+    @Param('id', ParseIntPipe) webhookId: number,
+  ): Promise<void> {
+    this.webhookService.deleteWebhook(userId, webhookId);
+  }
+
+  @UseGuards(JwtAuthGuard)
+  @ApiOperation({ summary: 'List all webhooks owned by user' })
+  @Get()
```
rad20c, line 47: if we'd like to strictly follow the convention then we should add `@HttpCode(200)` here as well. (It's by default 200 for get, 201 for post)

src/webhook/webhook.controller.ts
```diff
@@ -0,0 +1,51 @@
...
+  async listWebhooks(@UserId() userId: number): Promise<ListWebhooksDto[]> {
+    return await this.webhookService.getWebhooksById(userId);
```
rad20c, line 49: `getWebhooksByUserId`

src/webhook/webhook.controller.spec.ts
KubaBaniak: Mentioned this in comment above PR post. (tests are in next PR 😉)

src/webhook/dto/webhook-response.ts
```diff
@@ -0,0 +1,18 @@
+export interface FetchedWebhook {
+  id: number;
+  name: string;
+  url: string;
+  type: string;
+}
+
+export class FetchWebhooksResponse {
+  public fetchedWebhooks: FetchedWebhook[];
```
rad20c, line 9: hmm, this is in response right? I would just call it simply `webhooks`

src/webhook/webhook.controller.ts
```diff
@@ -0,0 +1,62 @@
...
+  @ApiOperation({ summary: 'Create webhook' })
+  @Post()
+  async createWebhook(
+    @UserId() userId: number,
+    @Body() webhookData: CreateWebhookRequest,
+  ): Promise<void> {
+    try {
+      await this.webhookService.createWebhook(userId, webhookData);
+    } catch {
+      throw new ForbiddenException(
```
rad20c, line 35: I would personally use 409 Conflict here, but 403 Forbidden technically seems correct too

rad20c: However, I think you should call `ForbiddenException` within `webhookService` where you are checking for `WEBHOOK_LIMIT`, not in controller

rad20c: Also, consider returning created webhook data - might be useful for the frontend. But it's also fine not to return any data.

src/webhook/webhook.service.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+      imports: [HttpModule],
+      providers: [WebhookService, WebhookRepository, PrismaService],
+    }).compile();
+
+    service = module.get<WebhookService>(WebhookService);
+  });
+
+  it('should be defined', () => {
+    expect(service).toBeDefined();
+  });
```
rad20c, line 21: would be good to add some tests

src/webhook/webhook.service.ts
```diff
@@ -0,0 +1,99 @@
...
+    this.webhookRepository.createWebhook(userId, webhookData);
+  }
+
+  async deleteWebhook(userId: number, webhookId: number) {
+    const webhook = await this.webhookRepository.getWebhookById(webhookId);
+    if (!webhook) {
+      throw new NotFoundException();
+    }
+    if (userId !== webhook.userId) {
+      throw new UnauthorizedException();
```
rad20c, line 46: Should be 403, not 401 in this case. Eventually if we want to be secretive about our resources - 404.

src/webhook/webhook.service.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  async createWebhook(
+    userId: number,
+    webhookData: CreateWebhookRequest,
+  ): Promise<void> {
+    const userWebhooks = await this.webhookRepository.getAllWebhooksByUserId(
+      userId,
+    );
+    if (userWebhooks.length >= +process.env.WEBHOOK_LIMIT) {
+      throw new ForbiddenException();
```
rad20c, line 29: see [comment here](https://github.com/KubaBaniak/RecipeAppNestjs/pull/48/files#r1305294275) - you should create the forbidden response here

src/webhook/webhook.service.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  async deleteWebhook(userId: number, webhookId: number) {
+    const webhook = await this.webhookRepository.getWebhookById(webhookId);
+    if (!webhook) {
+      throw new NotFoundException();
+    }
+    if (userId !== webhook.userId) {
+      throw new UnauthorizedException();
+    }
+    this.webhookRepository.deleteUserWebhookById(webhookId);
```
rad20c, line 48: missing `await`? (unless we want to return the response without waiting the operation to finish? But I wouldn't do that in this case)

src/webhook/webhook.service.ts
```diff
@@ -0,0 +1,99 @@
...
+      'Content-Type': 'application/json',
+      Authorization: `Bearer ${token}`,
+    };
+    this.httpService
+      .post(url, data, {
+        headers: headersRequest,
+      })
+      .subscribe({
+        error: () => {
+          const maxWaitingTime = 86400 * 1000;
```
rad20c, line 75: this is 24 hours in miliseconds right? I would write it like this: `1000 * 60 * 60 * 24` + move it to const value, e.g. `MAX_WAITING_TIME_IN_MS` (not sure about the overall naming, but note the unit at the end (ms))

src/webhook/webhook.service.ts
```diff
@@ -0,0 +1,99 @@
...
+          const nextTryInSec = Math.min(Math.pow(2, attempt + 1) * 1000);
```
rad20c, line 76: hmm, so your retry policy is 2s, 4s, 8s, 16s, ..., ~68 minutes, ..., ~18h 12m. I kind of don't like it 😛 Let's use our learning time and implement the following retry policy:

Retry | Time
1 | 2 minutes after the failure
2 | 6 minutes after the previous retry
3 | 30 minutes after the previous retry
4 | 1 hour after the previous retry
5 | 5 hours after the previous retry
6 | 1 day after the previous retry
7 | 2 days after the previous retry

(you can add it to the acceptance criteria)

src/webhook/webhook.service.ts
```diff
@@ -0,0 +1,99 @@
...
+          if (nextTryInSec > maxWaitingTime) {
+            throw new NotFoundException({
+              message: 'Could not send data to provided URL',
+            });
```
rad20c, line 80: I assume this is handled somehow (not crashing the app?)

src/webhook/webhook.service.ts
```diff
@@ -0,0 +1,99 @@
...
+          }
+          setTimeout(() => {
+            this.sendToWebhook(url, data, token, attempt + 1);
+          }, nextTryInSec);
```
rad20c, line 84: hmm, using `setTimeout` in this case might to be the best. Once there are more users in the system using webhooks, you will end up with a lot of `setTimeout` calls running in the memory that could affect backend performance. Also once you restart the server (e.g. because you are deploying a new version or an unexpected error occurred) you will lose all timeouts you have created.

I would change the approach here - use `cron` scheduler. You can run it e.g. every 1 minute and check if there are any webhook events that could be resent (only failed webhook events, new webhook events should be sent immediately, outside of cron).

Although using cron introduces a horizontal scaling problem (we want to run cron once - with 2 or more backend instances cron will be called once on each instance), we can fix it by either using semaphores (e.g. using Redlock) or moving cron logic to a separate micro service that will always have 1 instance only. For now - we don't have to care about this problem.

src/webhook/webhook.service.ts
```diff
@@ -0,0 +1,99 @@
...
+        },
+      });
+  }
+
+  async sendWebhookEvent(userId: number, data: Recipe, event: WebhookEvent) {
```
rad20c, line 89: missing return type

rad20c (changes requested):
see my comments

src/webhook/webhook.service.ts
```diff
@@ -0,0 +1,99 @@
...
+import { Recipe } from '@prisma/client';
+import Cryptr from 'cryptr';
+
+@Injectable()
+export class WebhookService {
+  constructor(
+    private readonly httpService: HttpService,
+    private readonly webhookRepository: WebhookRepository,
+  ) {}
+  cryptr = new Cryptr(process.env.CRYPTR_SECRET_KEY);
```
rad20c, line 19: Didn't know about this library - I see it's an easy to use library to use `aes-256-gcm` algorithm.

Normally I would say - ok 👍🏻 - but because we are learning, I would like you to use a built-in `node:crypto` library https://nodejs.org/api/crypto.html which can be used for AES and more. Hint #1: you might end up with an "util" class for this as there will be some code here. Hint #2: AES-256-GCM is a good choice 👍🏻 Don't use AES-256-CBC.

package.json
```diff
@@ -46,6 +48,7 @@
     "cache-manager-redis-store": "2.0.0",
     "class-transformer": "0.5.1",
     "class-validator": "0.14.0",
+    "cryptr": "6.2.0",
```
rad20c, line 51: you can remove the library now

prisma/migrations/20230901100610_add_iv_and_authtag_for_webhook_tokens/migration.sql
```diff
@@ -0,0 +1,3 @@
+-- AlterTable
+ALTER TABLE "Webhook" ADD COLUMN     "authTag" TEXT,
+ADD COLUMN     "initVector" TEXT;
```
rad20c, line 3: I think it should be in WebhookEvent

src/webhook/__mocks__/crypt-webhook-token.mock.ts
```diff
@@ -0,0 +1,20 @@
+import { faker } from '@faker-js/faker';
+import { TokenCrypt } from '../utils/crypt-webhook-token';
+
+export class MockTokenCrypt implements Required<TokenCrypt> {
+  encryptToken(_token: string): {
+    encryptedToken: string;
+    iv: string;
+    authTag: string;
```
rad20c, line 8: perhaps we could create a type / interface for this

src/webhook/utils/crypt-webhook-token.ts
```diff
@@ -0,0 +1,35 @@
+import crypto from 'crypto';
+import { Injectable } from '@nestjs/common';
+
+@Injectable()
+export class TokenCrypt {
```
rad20c, line 5: hmm, how about `CryptoUtils`?

src/webhook/utils/crypt-webhook-token.ts
```diff
@@ -0,0 +1,35 @@
...
+  private algorithm: crypto.CipherGCMTypes = 'aes-256-gcm';
+
+  encryptToken(token: string): {
+    encryptedToken: string;
+    iv: string;
+    authTag: string;
```
rad20c, line 11: as commented somewhere up, could extract it to a type / interface

src/webhook/utils/crypt-webhook-token.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  decryptToken(encryptedToken: string, iv: string, authTag: string): string {
+    const secretKey = Buffer.from(process.env.WEBHOOK_TOKEN_SECRET_KEY, 'hex');
+    const decipher = crypto.createDecipheriv(this.algorithm, secretKey, iv);
+    decipher.setAuthTag(Buffer.from(authTag, 'hex'));
+
+    let decryptedText = decipher.update(encryptedToken, 'hex', 'utf-8');
+
+    return (decryptedText += decipher.final('utf-8'));
+  }
```
rad20c, line 34: good job figuring this out 👍🏻

src/webhook/utils/retry-policy.ts
```diff
@@ -0,0 +1,10 @@
+//attempt: time in ms
+export const retryPolicy = {
+  1: 2 * 60 * 1000,
+  2: 6 * 60 * 1000,
+  3: 30 * 60 * 1000,
+  4: 1 * 60 * 60 * 1000,
+  5: 5 * 50 * 60 * 1000,
+  6: 1 * 24 * 60 * 60 * 1000,
+  7: 2 * 24 * 60 * 60 * 1000,
+};
```
rad20c, line 10: 👍🏻

src/webhook/webhook.repository.ts
```diff
@@ -0,0 +1,114 @@
...
+        userId,
+        initVector: iv,
+        authTag,
+      },
+    });
+  }
+
+  async createWebhookEvent(
+    webhookId: number,
+    data: any,
```
rad20c, line 28: perhaps `unknown` would be better (not always possible to use)

src/webhook/webhook.service.ts
```diff
@@ -0,0 +1,99 @@
...
+    const headersRequest = {
+      'Content-Type': 'application/json',
+      Authorization: `Bearer ${token}`,
+    };
+    const test = await firstValueFrom(
+      this.httpService.post(url, data, {
+        headers: headersRequest,
+      }),
+    );
+    return test.status === 200 ? true : false;
```
rad20c, line 83: `return test.status === 200` should be enough, no need for `true : false`

src/webhook/webhook.service.ts
```diff
@@ -0,0 +1,99 @@
...
+    userId: number,
+    data: Recipe,
+    type: WebhookEventType,
+  ): Promise<void> {
+    const userWebhooks = await this.webhookRepository.getAllWebhooksByUserId(
+      userId,
+    );
+    userWebhooks.forEach(async (webhook) => {
+      if (webhook.type === type)
+        await this.webhookRepository.createWebhookEvent(webhook.id, data, type);
```
rad20c, line 96: Personally I would avoid one line ifs - it might be confusing if you later will have to add a new line under this `await` and you will be wondering why it's being called every time instead of only when the condition is matched.

In short: always use curly braces:
```suggestion
      if (webhook.type === type) {
        await this.webhookRepository.createWebhookEvent(webhook.id, data, type);
      }
```

rad20c (approved):
good job 👍🏻

PR #49, opened 2023-08-21
test: webhooks

KubaBaniak:
Added unit & e2e tests for webhooks


src/webhook/test/webhook.controller.spec.ts
```diff
@@ -22,9 +26,98 @@ describe('WebhookController', () => {
...
+      };
+
+      controller.createWebhook(userId, data);
+
+      expect(webhookService.createWebhook).toHaveBeenCalled();
+    });
+  });
+
+  describe('Delete webhook', () => {
+    it('should delete a token given id', () => {
```
rad20c, line 79: it looks like a copied test name from other test file - update the test name

src/webhook/test/webhook.controller.spec.ts
```diff
@@ -22,9 +26,98 @@ describe('WebhookController', () => {
...
+        );
+
+      const userId = faker.number.int();
+      const webhookId = faker.number.int();
+
+      controller.deleteWebhook(userId, webhookId);
+
+      expect(webhookService.deleteWebhook).toHaveBeenCalled();
+
+      jest.clearAllMocks();
```
rad20c, line 93: It's safer to move `jest.clearAllMocks();` to `afterEach` method. Imagine if `expect(webhookService.deleteWebhook).toHaveBeenCalled();` fails in this test - `jest.clearAllMocks();` won't be called and other tests might be affected

src/webhook/test/webhook.factory.ts
```diff
@@ -0,0 +1,46 @@
...
+  url: overrides.url ?? faker.internet.url(),
+  token: overrides.token ?? faker.string.alphanumeric(32),
+  userId,
+  initVector: faker.string.alphanumeric(16),
+  authTag: faker.string.alphanumeric(16),
+});
+
+export function createWebhookResponse() {
+  return {
+    id: faker.number.int({ max: 128 }),
```
rad20c, line 37: hmm, in `createWebhookWithUserId()` method you have `id: faker.number.int({ max: 65535 })` and in `createWebhookResponse()` `id: faker.number.int({ max: 128 })`. What are these `128` and `65535` values?

src/webhook/webhook.service.ts
```diff
@@ -43,18 +43,18 @@ export class WebhookService {
...
+  async deleteWebhook(userId: number, webhookId: number): Promise<void> {
     const webhook = await this.webhookRepository.getWebhookById(webhookId);
     if (!webhook) {
       throw new NotFoundException();
     }
     if (userId !== webhook.userId) {
       throw new ForbiddenException();
     }
-    return this.webhookRepository.deleteUserWebhookById(webhookId);
+    await this.webhookRepository.deleteUserWebhookById(webhookId);
```
rad20c, line 57: Probably you don't need await here - would have to check

test/webhook.e2e-spec.ts
```diff
@@ -0,0 +1,141 @@
...
+        .expect(HttpStatus.CREATED);
+      expect(
+        await prismaService.webhook.findFirst({
+          where: { name: webhook.name },
+        }),
+      ).toBeDefined();
+    });
+    it('should not create new webhook due to the limit (5 webhooks)', async () => {
+      const testWebhooks = [
+        createWebhookWithUserId(user.id, {}),
```
rad20c, line 78: I think the second parameter `{}` is not necessary as in the `createWebhookWithUserId` definition you have `= {}` which sets a default value in case the param is missing.
```suggestion
        createWebhookWithUserId(user.id),
```

rad20c (approved):

src/webhook/test/webhook.factory.ts
KubaBaniak: There is an error, when faker generates too big Integers. 
```
"Unable to fit integer value '3313120212680704' into an INT4 (32-bit signed integer)."
```
so I hard coded those values to 'less bit' integers

PR #52, opened 2023-08-24
chore: send and use sourcemaps

KubaBaniak:


nest-cli.json
```diff
@@ -1,7 +1,6 @@
 {
   "$schema": "https://json.schemastore.org/nest-cli",
   "collection": "@nestjs/schematics",
-  "sourceRoot": "src",
```
rad20c, line 4: are you sure about removing `sourceRoot`? Why did you have to remove it?

Dockerfile.nginx
rad20c, line 1: As discussed on our call - I think you don't need to create a separate Dockerfile for uploading source maps purposes and instead can utilize multi-stage targeting as described [here](https://reece.tech/posts/extracting-files-multi-stage-docker/) or [here](https://stackoverflow.com/questions/51253987/building-a-multi-stage-dockerfile-with-target-flag-builds-all-stages-instead-o).

rad20c (approved):

PR #53, opened 2023-09-06
feat: account activation with email

KubaBaniak:
### Acceptance criteria:
- User has to activate account to access other endpoints
- User receives an email with activation link
- If User did not activate account in given time - provided data is deleted
- Token is valid for 24h


src/mail/mail.service.ts
```diff
@@ -0,0 +1,21 @@
...
+
+@Injectable()
+export class MailService {
+  constructor(private mailerService: MailerService) {}
+
+  async sendAccountActivationEmail(
+    userEmail: string,
+    token: string,
+  ): Promise<SentMessageInfo> {
+    const url = `http://localhost:3000/auth/activate-account/?token=${token}`;
```
PrzerostNadSens, line 13: change it to `baseUrl` from env 
```suggestion
    const url = `${baseUrl}/auth/activate-account/?token=${token}`;
```

PrzerostNadSens (changes requested):

src/auth/auth.service.ts
```diff
@@ -16,33 +17,58 @@ import {
...
 export class AuthService {
   constructor(
     private readonly userRepository: UserRepository,
     private readonly personalAccessTokenRepository: PersonalAccessTokenRepository,
     private readonly jwtService: JwtService,
+    private readonly accountActivationTimeouts: AccountActivationTimeouts,
   ) {}
 
-  async verifyJwt(jwtToken: string): Promise<AccessTokenPayload> {
+  verifyJwt(jwtToken: string): Promise<AccessTokenPayload> {
```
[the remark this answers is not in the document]
KubaBaniak: Can't remove `Promise<...>`, because inside I return `Promise<any>` from `jwtService.verifyAsync()`

example.env
```diff
@@ -1,20 +1,28 @@
...
+ENVIRONMENT=
+JWT_ACCOUNT_ACTIVATION_SECRET=
 JWT_EXPIRY_TIME=
-JWT_SECRET=
 JWT_PAT_SECRET=
-BCRYPT_SALT_ROUNDS=
+JWT_SECRET=
+MAIL_FROM=
+MAIL_HOST=
+MAIL_PASSW
```
rad20c, line 15: hmm, something went wrong here 😅

prisma/schema.prisma
```diff
@@ -11,13 +11,15 @@ datasource db {
...
-  role                Role                  @default(USER)
-  recipes             Recipe[]
-  webhooks            Webhook[]
-  personalAccessToken PersonalAccessToken[]
+  id                     Int                   @id @default(autoincrement())
+  email                  String                @unique
+  password               String
+  role                   Role                  @default(USER)
+  activated              Boolean               @default(false)
+  accountActivationToken String?
```
rad20c, line 19: I would move it to a separate table.

Before user activates his/her account, there should be no `User` record created (no account should be created yet).

We could however assume that only `activated === true` accounts are created / active accounts, but then when developing other features related to user, e.g. "invite user to organisation" we would have to remember not to do a simple `select * from User` but add a `where` clause where we would filter out not activated accounts. What I'm trying to say - moving this to a separate table - although not necessary - might save us some trouble later.

After activation we can either remove the record from the new table or leave the record but keep only the necessary data (so e.g. remove the password).

Also, I would rename `activated` to `isActivated`

src/auth/utils/timeout-functions.ts
```diff
@@ -0,0 +1,36 @@
...
+    this.schedulerRegistry.addTimeout(name, timeout);
+  }
+
+  deleteTimeout(name: string): void {
+    this.schedulerRegistry.deleteTimeout(name);
+  }
+
+  getAllTimeouts(): string[] {
+    return this.schedulerRegistry.getTimeouts();
+  }
```
rad20c, line 35: If this is using `.setTimeout` under the hood then I think we should not use this. Remember our discussion about what happens when you restart the service? Unless I'm missing something and you have addressed that. But I think we should use cron instead. I assume that the account activation time is somewhere around 24h.

src/auth/auth.service.ts
```diff
@@ -102,6 +128,50 @@ export class AuthService {
...
+    }
+  }
+
+  async activateAccount(userId: number): Promise<User> {
+    const timeoutName = this.accountActivationTimeouts.getName(userId);
+    try {
+      this.accountActivationTimeouts.deleteTimeout(timeoutName);
+    } catch {
+      throw new NotFoundException(
+        `Account deletion timeout has been already deleted, therefore your accound is already activated.`,
```
rad20c, line 169: As a user I have no idea what it means 😅 I even have trouble understanding it as a developer 😅 

Why "Account deletion timeout" deleted => account activated? What if account was not activated in time and it has been deleted? Or does it not work this way?

rad20c: also, `therefore your accound` -> `therefore your account` - always double check the PRs! :)

rad20c (changes requested):
Check comments

src/auth/auth.service.ts
```diff
@@ -25,24 +25,57 @@ export class AuthService {
...
+  async generateBearerToken(
+    id: number,
+    secret: string,
+    timeInSeconds?: number,
+  ): Promise<string> {
+    const payload = { id };
+    const options: JwtSignOptions = { secret };
+
+    if (timeInSeconds) {
+      options.expiresIn = `${timeInSeconds}s`;
```
rad20c, line 41: hmm, perhaps I wouldn't hardcode seconds here, but just pass a value from environment value that would be either `60s` or `2h` or `7d` or something else

src/user/user.repository.ts
```diff
@@ -42,9 +39,66 @@ export class UserRepository {
...
+  ): Promise<{ accountActivationToken: string }> {
+    return this.prisma.pendingUser.findUnique({
+      where: {
+        id: userId,
+      },
+      select: {
+        accountActivationToken: true,
+      },
+    });
+  }
```
rad20c, line 103: I would create a separate repository for all these methods

prisma/schema.prisma
```diff
@@ -20,6 +20,14 @@ model User {
   personalAccessToken PersonalAccessToken[]
 }
 
+model PendingUser {
+  id                     Int      @id @default(autoincrement())
+  email                  String   @unique
+  password               String
+  accountActivationToken String?
```
rad20c, line 27: hmm, if we have decided to store the activation token, why is it optional?

I'm not sure if we need to store such token though in the database

src/auth/auth.service.ts
```diff
@@ -25,24 +25,57 @@ export class AuthService {
...
+
+    return this.jwtService.signAsync(payload, options);
+  }
+
+  async signUp(signUpRequest: SignUpRequest): Promise<SignUpResponse> {
+    const pendingUser = await this.userRepository.getPendingUserByEmail(
+      signUpRequest.email,
+    );
+
+    const user = await this.userRepository.getUserByEmail(signUpRequest.email);
```
rad20c, line 52: these can be fetched in parallel using `Promise.all`

rad20c, line 52: these can be fetched in parallel using `Promise.all`

src/auth/auth.service.ts
```diff
@@ -102,6 +114,37 @@ export class AuthService {
...
+      throw new ForbiddenException('Token expired');
+    }
+  }
+
+  async activateAccount(userId: number): Promise<UserPayloadRequest> {
+    const userData = await this.userRepository.getPendingUserById(userId);
+
+    await this.userRepository.removePendingUserById(userId);
+
+    return this.userRepository.createUser(userData);
```
rad20c, line 145: hmm, not that this is wrong, but I'd rather first create resources, then remove old ones.

Also a note (you don't have to do this) - using a raw SQL you should probably be able to copy data from pendingUser to user without loading them into the backend

src/auth/test/auth.service.spec.ts
```diff
@@ -58,18 +54,21 @@ describe('AuthService', () => {
...
-      jest
-        .spyOn(userRepository, 'getUserByEmailWithPassword')
-        .mockImplementation((email) => {
-          return Promise.resolve({
-            id: faker.number.int(),
-            email,
-            password: request.password,
-            role: Role.USER,
-          });
+      const hashed_password = await bcrypt.hash(
```
rad20c, line 57: why not camelCase?
```suggestion
      const hashedPassword = await bcrypt.hash(
```

src/auth/test/auth.service.spec.ts
```diff
@@ -132,6 +131,32 @@ describe('AuthService', () => {
     });
   });
 
+  describe('Generace account activation token', () => {
+    it('should generate and save token for account activation.', async () => {
+      //given
+      const userId = faker.number.int({ max: 2147483647 });
```
rad20c, line 137: Perhaps you could move `2147483647` to const and name it as a max int32 number

src/prisma/__mocks__/prisma.service.mock.ts
```diff
@@ -48,6 +58,61 @@ export class MockPrismaService {
...
+    },
+
+    findUnique: function (request: {
+      where: { id?: number; email?: string };
+    }): Promise<PendingUser> | Promise<void> {
+      const { id, email } = request.where;
+
+      if (email) {
+        return Promise.resolve();
+      }
```
rad20c, line 109: hmm, I don't quite understand this mock - if `email` is present then we don't return anything?

src/user/user.module.ts
```diff
@@ -3,10 +3,11 @@ import { UserController } from './user.controller';
 import { UserService } from './user.service';
 import { PrismaService } from '../prisma/prisma.service';
 import { UserRepository } from './user.repository';
+import { PendingUserDeletion } from './pending-user-cron';
 
 @Module({
   controllers: [UserController],
-  providers: [UserService, UserRepository, PrismaService],
+  providers: [UserService, UserRepository, PrismaService, PendingUserDeletion],
```
rad20c, line 10: the name does not say what this class is exactly. It should be e.g. `PendingUserDeletionCronService`

src/user/test/user.service.spec.ts
```diff
@@ -38,7 +37,7 @@ describe('UserService', () => {
       expect(createdUser).toEqual({
         id: expect.any(Number),
         email: request.email,
-        role: Role.USER,
+        password: undefined,
```
rad20c, line 40: why do we check for `undefined` password?

test/auth.e2e-spec.ts
```diff
@@ -69,8 +77,16 @@ describe('AuthController (e2e)', () => {
   });
 
   describe('POST /auth/signin', () => {
+    let user: { email: string; password: string };
     beforeEach(async () => {
-      await authService.signUp(user);
+      user = createUser();
+      const hashed_password = await bcrypt.hash(
```
rad20c, line 83:
```suggestion
      const hashedPassword = await bcrypt.hash(
```

test/auth.e2e-spec.ts
```diff
@@ -108,10 +124,33 @@ describe('AuthController (e2e)', () => {
...
+      const createdUser = await authService.signUp({
+        email: user.email,
+        password: user.password,
+      });
+      token = await authService.generateAccountActivationToken(createdUser.id);
+    });
+
+    it('should activate an account', async () => {
+      return request(app.getHttpServer())
+        .get(`/ auth / activate - account /? token = ${token}`)
```
rad20c, line 140: hmm, that's the first time I'm seeing such a thing 🤔 Is it the same as:

```
.get(`/auth/activate-account/?token=${token}`)
```
? Does it work?

Still, I think i would be better to omit the spaces so that it's not mistaken as some kind of comment 😛

rad20c (approved):

src/prisma/__mocks__/prisma.service.mock.ts
KubaBaniak: Yea, I use this if to be able to get correct response on `signUp` tests - if this `findUnique({where: {email: ... }})` would return something, then user would get 401

PR #55, opened 2023-09-07
feat: 2FA

KubaBaniak:
# ⚠️ Tests will be included in next PR ⚠️
### Acceptance criteria:
- User can turn on and off 2FA
- After user logged in we require him to provide TOTP (Time-based one-time password) - only if user has enabled 2FA.
- User has been provided recovery tokens During 2FA set up process.


example.env
```diff
@@ -18,3 +18,6 @@ WEBHOOK_LIMIT=
 ROLLBAR_ACCESS_TOKEN=
 ENVIRONMENT=
 WEBHOOK_TOKEN_SECRET_KEY=
+
+SECRET_KEY_2FA= // this has to be base32 to work with Google Authenticator
```
rad20c, line 22: hmm, I wouldn't add comments to `example.env`. If anywhere, I'd put it in README.md

prisma/schema.prisma
```diff
@@ -18,6 +18,8 @@ model User {
   recipes             Recipe[]
   webhooks            Webhook[]
   personalAccessToken PersonalAccessToken[]
+  enabled2FA          Boolean               @default(false)
+  recoveryKeys        String[]
```
rad20c, line 22: I would also add a 2fa type column - currently it can be set to e.g. "authenticator" value by default. It will allow us to differentiate between other 2fa types (like SMS) if want to add extra 2fa type in the future.

rad20c: Also I would consider moving 2fa related fields (and perhaps user credentials too?) to a separate table with 1:1 relation. `recoveryKeys` especially are not obvious what are they related to.

rad20c: One more thing: I'm missing 2fa secret. How are we going to validate the user token without knowing the secret?

src/auth/__mocks__/auth.controller.mock.ts
```diff
@@ -19,6 +19,7 @@ export class MockAuthController {
       id: faker.number.int(),
       email: signUpRequest.email,
       password: signUpRequest.password,
+      enabled2FA: false,
```
rad20c, line 22: this name kind of doesn't sit with me - maybe `secondFactorEnabled` or `mfaEnabled`. Something else so that the `enabled` part is at the end

src/auth/auth.controller.ts
```diff
@@ -75,4 +75,77 @@ export class AuthController {
       changePasswordRequest.newPassword,
     );
   }
+
+  @HttpCode(201)
+  @ApiOperation({
+    summary:
+      'Creates QR code for user to scan it for auth app (like google Authenticator)',
```
rad20c, line 82:
```suggestion
      'Creates QR code for user to scan it for auth app (like Google Authenticator)',
```

src/auth/dto/create-qrcode-2fa-response.ts
```diff
@@ -0,0 +1,12 @@
+export class CreateQrcodeFor2FA {
+  public qrcodeUrl: string;
+  public urlToEnable2FA = 'http://localhost:3000/api/auth/enable-2fa';
```
rad20c, line 3: `http://localhost:3000` part should be read from environment variable. The current one will not work on staging.

src/auth/auth.controller.ts
```diff
@@ -75,4 +75,77 @@ export class AuthController {
...
+  })
+  @UseGuards(JwtAuthGuard)
+  @Post('create-qr-2fa')
+  async createQrcodeFor2FA(
+    @UserId() userId: number,
+  ): Promise<CreateQrcodeFor2FA> {
+    const qrcode = await this.authService.createQrcodeFor2FA(userId);
+
+    return CreateQrcodeFor2FA.from(qrcode);
```
rad20c, line 91: hmm ok, for learning purposes we can leave the logic like this 👍🏻

If it was a production app I would leave the QR code generation in the frontend app hands, and here I would return a generated secret (some flows allow you to manually type secret to authenticator app instead of scanning a QR code)

src/auth/auth.controller.ts
```diff
@@ -75,4 +75,77 @@ export class AuthController {
...
+    return Enable2FARespnse.from(recoveryKeys);
+  }
+
+  @HttpCode(200)
+  @ApiOperation({ summary: 'Disables 2FA for logged user' })
+  @UseGuards(JwtAuthGuard)
+  @Post('disable-2fa')
+  disable2FA(@UserId() userId: number): void {
+    this.authService.disable2FA(userId);
+  }
```
rad20c, line 118: I wonder if these two endpoints should be in `user` module instead of `auth` module? That is if we treat `auth` module as a module containing authentication / authorization methods. If you want to treat `auth module as module containing everything related to authentication / authorization (so including password reset, email change etc.) then I guess it's ok.

src/auth/auth.service.ts
```diff
@@ -29,6 +31,16 @@ export class AuthService {
     return this.jwtService.verifyAsync(jwtToken);
   }
 
+  async successfullLoginToken(id: number, email: string): Promise<string> {
```
rad20c, line 34: how about something like `generateUserBearerToken`? You could include strategy name in the method name but I don't remember the strategy names you have.

src/auth/auth.service.ts
```diff
@@ -37,13 +49,19 @@ export class AuthService {
...
-      },
-      { expiresIn: `${process.env.JWT_EXPIRY_TIME}s` },
-    );
+    if (user.enabled2FA) {
+      return this.jwtService.signAsync(
+        {
+          id: user.id,
+          status: 'FURTHER_ACTION_IS_REQUIRED',
+          message:
+            'Go to http://localhost:3000/auth/verify-2fa to continue authentication',
```
rad20c, line 58: `http://localhost:3000` part should be taken from environment variables. Otherwise it won't work on environments different from development / local one.

src/auth/auth.service.ts
```diff
@@ -112,4 +130,60 @@ export class AuthService {
       password: hashedPassword,
     });
   }
+
+  async createQrcodeFor2FA(userId: number): Promise<string> {
+    const email = (await this.userRepository.getUserById(userId)).email;
```
rad20c, line 135:
```suggestion
    const { email } = await this.userRepository.getUserById(userId);
```

src/auth/auth.service.ts
```diff
@@ -112,4 +130,60 @@ export class AuthService {
...
+    return qrcode.toDataURL(otpauth);
+  }
+
+  async enable2FA(userId: number, providedToken: string): Promise<string[]> {
+    const user = await this.userRepository.getUserById(userId);
+
+    if (authenticator.check(providedToken, process.env.SECRET_KEY_2FA)) {
+      const recoveryKeys: string[] = Array.from({ length: 3 }, () =>
+        authenticator.generateSecret(),
+      );
```
rad20c, line 153: I think you should generate at least 8 codes

src/auth/auth.service.ts
rad20c: seeing that this is a token payload not a HTTP response, I'm not sure if `status` and `message` are necessary - I would remove them from the token payload I guess

src/auth/auth.service.ts
```diff
@@ -112,4 +130,60 @@ export class AuthService {
       password: hashedPassword,
     });
   }
+
+  async createQrcodeFor2FA(userId: number): Promise<string> {
+    const email = (await this.userRepository.getUserById(userId)).email;
+    const service = 'Recipe App';
```
rad20c, line 136: candidate to be moved to some global consts. Though currently I don't see it being used anywhere else.

src/auth/auth.service.ts
```diff
@@ -112,4 +130,60 @@ export class AuthService {
...
+  ): Promise<string> {
+    const { recoveryKeys, email } =
+      await this.userRepository.get2FARecoveryKeysAndEmailByUserId(userId);
+
+    if (!recoveryKeys.includes(providedRecoveryKey)) {
+      throw new ForbiddenException();
+    }
+
+    return this.successfullLoginToken(userId, email);
+  }
```
rad20c, line 188: After the recovery code is used, it should be marked as used and should no longer be available for use. Consider also an endpoint for regenerating the recovery codes.

src/auth/dto/index.ts
```diff
@@ -6,3 +6,7 @@ export * from './user-request';
 export * from './create-pat-response';
 export * from './access-token-payload';
 export * from './change-password-request';
+export * from './create-qrcode-2fa-response';
+export * from './verify-2fa-request';
+export * from './recovery-2fa-request';
+export * from './enable-2fa-response';
```
rad20c, line 12: if you have a vscode for organizing imports then use it to have imports in alphabetical order 🙂

src/auth/dto/sign-up-response.ts
```diff
@@ -1,9 +1,14 @@
...
+  constructor(
+    public id: number,
+    public email: string,
+    public role: string,
+    public enabled2FA: boolean,
+  ) {}
 
   public static from(user: UserPayloadRequest): SignUpResponse {
-    return new SignUpResponse(user.id, user.email, user.role);
+    return new SignUpResponse(user.id, user.email, user.role, user.enabled2FA);
```
rad20c, line 12: do we need to return `user.enabled2FA` after creating an account?

src/auth/guards/two-factor-auth.guard.ts
```diff
@@ -0,0 +1,5 @@
+import { Injectable } from '@nestjs/common';
+import { AuthGuard } from '@nestjs/passport';
+
+@Injectable()
+export class TwoFactorAuthGuard extends AuthGuard('2FA.bearer') {}
```
rad20c, line 5: hmm, to be honest I don't know if there are any naming conventions for auth guards, but I guess I would change it to either: `2fa.bearer` or `2fa-bearer`

src/auth/strategies/jwt.strategy.ts
```diff
@@ -15,7 +15,10 @@ export class UserAuthBearerStrategy extends PassportStrategy(
       secretOrKey: jwtConstants.secret,
     });
   }
-  validate(payload: JwtPayload): JwtPayload {
+  validate(payload: any): JwtPayload {
+    if (payload.status === 'FURTHER_ACTION_IS_REQUIRED') {
+      throw new ForbiddenException(payload.message);
+    }
```
rad20c, line 21: hmm I don't understand this method - it says `validate` but we are checking the payload. We should not rely on payload for validation but on the JWT signature that every JWT token should have.

Also `JwtPayload` changed to `any` - I have a hunch that you are using this method for both bearer jwt token and 2fa bearer token. Don't do it like this. We should have a separate validate method for each JWT type. For each we should try and validate them using their signatures. If the signature is invalid then obviously it's not a valid token of this type.

src/auth/strategies/payloads/two-factor-auth.payload.ts
```diff
@@ -0,0 +1,5 @@
+export interface TwoFactorAuthPayload {
+  id: number;
+  status: string;
+  message: string;
```
rad20c, line 4: as I said in some other comment, I would remove `status` and `message` from the token payload

src/auth/strategies/two-factor-auth.strategy.ts
```diff
@@ -0,0 +1,21 @@
+import { ExtractJwt, Strategy } from 'passport-jwt';
+import { PassportStrategy } from '@nestjs/passport';
+import { Injectable } from '@nestjs/common';
+import { jwtConstants } from '../constants';
+import { TwoFactorAuthPayload } from './payloads/two-factor-auth.payload';
+
+@Injectable()
+export class TwoFactorAuthStrategy extends PassportStrategy(
+  Strategy,
+  '2FA.bearer',
```
rad20c, line 10: I see this string for a second time - move it to some more globally scoped constant

src/auth/strategies/two-factor-auth.strategy.ts
```diff
@@ -0,0 +1,21 @@
...
+) {
+  constructor() {
+    super({
+      jwtFromRequest: ExtractJwt.fromAuthHeaderAsBearerToken(),
+      secretOrKey: jwtConstants.secret,
+    });
+  }
+  validate(payload: TwoFactorAuthPayload): { id: number } {
+    return { id: payload.id };
+  }
```
rad20c, line 20: again, wes should validate not the payload, but the token signature

src/user/test/user.service.spec.ts
```diff
@@ -38,7 +38,9 @@ describe('UserService', () => {
       expect(createdUser).toEqual({
         id: expect.any(Number),
         email: request.email,
+        enabled2FA: false,
         role: Role.USER,
+        password: undefined,
```
rad20c, line 43: I think you don't have to add a `password: undefined`. You could consider `. toDeepEqual` method.

src/user/user.repository.ts
```diff
@@ -47,4 +47,44 @@ export class UserRepository {
       where: { id },
     });
   }
+
+  async enable2FAForUserWithId(
```
rad20c, line 51:
```suggestion
  async enable2FAForUserId(
```
?

rad20c (changes requested):
see comments

src/auth/dto/index.ts
KubaBaniak: Capture-2023-09-13-112256

KubaBaniak: I will find the reason why my editor doesn't apply auto-sorting and fix it in my free time 😉

src/auth/auth.controller.ts
KubaBaniak: I will treat this module as everything related to auth 😅

example.env
```diff
@@ -18,3 +18,6 @@ WEBHOOK_LIMIT=
 ROLLBAR_ACCESS_TOKEN=
 ENVIRONMENT=
 WEBHOOK_TOKEN_SECRET_KEY=
+SECRET_KEY_2FA=
+JWT_2FA_EXPIRY_TIME=
+URL=
```
rad20c, line 23: url for what? should be better named

src/auth/auth.service.ts
```diff
@@ -29,21 +33,38 @@ export class AuthService {
...
-      signInRequest.email,
+  async generateBearerToken(id: number, email: string): Promise<string> {
+    return this.jwtService.signAsync(
+      {
+        id,
+        email,
+      },
+      {
+        secret: process.env.JWT_SECRET,
+        expiresIn: `${process.env.JWT_EXPIRY_TIME}s`,
```
rad20c, line 44: I wouldn't hardcode a limit to seconds. I would set values in envs like `60s` or `2h` or `7d` etc.

src/auth/auth.service.ts
```diff
@@ -29,21 +33,38 @@ export class AuthService {
...
-      { expiresIn: `${process.env.JWT_EXPIRY_TIME}s` },
-    );
+    if (user.twoFactorAuth) {
+      return this.jwtService.signAsync(
+        {
+          id: user.id,
+        },
+        {
+          secret: process.env.JWT_2FA_SECRET,
+          expiresIn: `${process.env.JWT_2FA_EXPIRY_TIME}s`,
```
rad20c, line 62: same as above

src/auth/constants.ts
```diff
@@ -1,9 +1,16 @@
...
+export const serviceConstants = { name: 'Recipe App' };
+
+export const strategyNameConstants = {
+  jwt: {
+    bearer: 'jwt.bearer',
+    pat: 'jwt.pat',
+    twoFactor: '2fa.bearer',
+  },
+  local: 'local',
+};
```
rad20c, line 16: check comment from other PR where I recommend changing the naming convention here

src/auth/constants.ts
```diff
@@ -1,9 +1,16 @@
...
   salt: +process.env.BCRYPT_SALT_ROUNDS,
 };
+
+export const serviceConstants = { name: 'Recipe App' };
+
+export const strategyNameConstants = {
+  jwt: {
+    bearer: 'jwt.bearer',
+    pat: 'jwt.pat',
+    twoFactor: '2fa.bearer',
```
rad20c, line 13: I would keep a flat structure here - not nest it in `jwt`. Especially since `twoFactor` does not belong under `jwt`

src/auth/dto/index.ts
```diff
@@ -6,3 +6,6 @@ export * from './user-request';
 export * from './create-pat-response';
 export * from './access-token-payload';
 export * from './change-password-request';
+export * from './create-qrcode-2fa-response';
+export * from './verify-2fa-request';
+export * from './recovery-keys-response';
```
rad20c, line 11: perhaps not that important, but it would be good to keep imports in alphabetical order. I wonder if there is some plugin for vscode that will automatically rearrange the imports 🤔 Perhaps typescript-hero or something similar

rad20c (approved):

rad20c (approved):

PR #56, opened 2023-09-08
test: 2FA

KubaBaniak:
# ⚠️ Tests for https://github.com/KubaBaniak/RecipeAppNestjs/pull/55


src/auth/constants.ts
```diff
@@ -6,6 +6,8 @@ export const bcryptConstants = {
 
 export const serviceConstants = { name: 'Recipe App' };
 
+export const numberOf2faRecoveryTokens = 8;
+
 export const strategyNameConstants = {
```
rad20c, line 11: hmm, I think we should change the naming convention for these values to Upper case letters. So:
```suggestion
export const SERVICE_CONSTANTS = { NAME: 'Recipe App' };

export const NUMBER_OF_2FA_RECOVERY_TOKENS = 8;

export const STRATEGY_NAME_CONSTANTS = {
```
etc.

also not sure about e.g. `STRATEGY_NAME_CONSTANTS`. Maybe it should be just a `STRATEGY` so that in other parts of the code we can use e.g. `STRATEGY.LOCAL`

src/auth/test/auth.controller.spec.ts
```diff
@@ -89,4 +90,54 @@ describe('AuthController', () => {
...
+      const userId = faker.number.int();
+      const tokenData = { token: faker.string.numeric(6) };
+
+      const responseObject = await authController.enable2FA(userId, tokenData);
+
+      expect(responseObject.recoveryKeys).toHaveLength(
+        numberOf2faRecoveryTokens,
+      );
+      responseObject.recoveryKeys.forEach((key) => {
+        expect(typeof key).toBe('string');
```
rad20c, line 117: I'd check if the `key`s are not empty

test/auth.e2e-spec.ts
```diff
@@ -32,52 +38,55 @@ describe('AuthController (e2e)', () => {
...
     });
 
     it('should not register a user (already in db) and return 403 error (FORBIDDEN ACCESS)', async () => {
-      await prismaService.user.create({ data: user });
       return request(app.getHttpServer())
         .post('/auth/signup')
         .set('Accept', 'application/json')
-        .send(user)
+        .send(tempUser)
         .expect(HttpStatus.FORBIDDEN);
```
rad20c, line 74: hmm, kind of risky - imagine if someone puts another test in between those two that will remove `tempUser`. Also - not sure - but I think that if you run a second test without running the first one, it will fail. And thus it violates the principle that tests should be deterministic.

You should register a user in the first part of the test (just don't put too many expects as you are doing this in other test - just one to check if the user was indeed registered is enough)

rad20c (approved):

PR #62, opened 2023-10-10
fix: account activation link and token

KubaBaniak:
### FIX: 
- Code 500 when we send an invalid token value or no token at all. The error should be handled.

- Email has a malformed activation link "undefined/auth/activate-account/"

### DIFF
--- a/src/auth/auth.service.ts
+++ b/src/auth/auth.service.ts
@@ -2,6 +2,7 @@ import {
   BadRequestException,
   ForbiddenException,
   Injectable,
+  NotFoundException,
   UnauthorizedException,
 } from '@nestjs/common';
 import { JwtService, JwtSignOptions } from '@nestjs/jwt';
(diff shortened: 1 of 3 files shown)


src/auth/auth.service.ts
```diff
@@ -121,19 +123,33 @@ export class AuthService {
   async verifyAccountActivationToken(
     jwtToken: string,
   ): Promise<{ id: number }> {
+    const invalidTokenMessage =
+      'Invalid token. Please provide a valid token to activate account';
+
+    if (!isJwtTokenStructureValid(jwtToken)) {
+      throw new UnauthorizedException(invalidTokenMessage);
+    }
```
rad20c, line 131: Do we need this check? `jwtService.verifyAsync` should be enough as it also checks the token structure

rad20c (approved):

PR #64, opened 2023-10-12
feat: allow users to create webhooks with multiple event types

KubaBaniak:
## FIX:
- We should be able to assign more than 1 event to a webhook

- We should also be able to tell which event triggered it

- Right now, POST /webhook properties such as name have no limit on max length. There should be one.


rad20c:
comment regarding PR title - this is not a `refactor:`, this sounds like a `feat:`

src/webhook/dto/create-webhook-request.ts
```diff
@@ -1,23 +1,30 @@
...
-import { IsIn, IsNotEmpty, IsOptional, IsString } from 'class-validator';
+import { IsOptional, IsString, IsUrl, Length } from 'class-validator';
+import { NoDuplicateAndOtherTypes } from './custom-class-validators';
 import { WebhookEventType } from './webhook-event-types';
 
 export class CreateWebhookRequest {
-  @IsNotEmpty()
   @IsString()
+  @Length(4, 255, {
+    message: 'name should should have 4-255 characters',
```
rad20c, line 9: hmm 
```suggestion
    message: 'name should should be between 4 and 255 characters long',
```

src/webhook/dto/create-webhook-request.ts
```diff
@@ -1,23 +1,30 @@
...
-  @IsNotEmpty()
-  @IsString()
+  @NoDuplicateAndOtherTypes(
+    ['recipe_created', 'recipe_updated', 'recipe_deleted'],
+    {
+      message: `Invalid or duplicate event type detected in the array. Use these types: ${Object.values(
+        WebhookEventType,
+      )}`,
+    },
+  )
```
rad20c, line 21: hmm, I wouldn't do it like this.

First problem is that it's doing two things instead of one. It should either check only for a duplicated entries or check only the type, not both at once. For checking the type we can use:
```typescript
@IsArray()
@IsString({ each: true })
```

for the duplicates, I would handle it another way - I would just sanitize the input - that is I would remove the duplicated entries from the array. You can use `Transform` from `class-transformer` for sanitization purposes (https://github.com/typestack/class-transformer and https://www.npmjs.com/package/@nestjs/class-transformer).

src/webhook/dto/custom-class-validators.ts
```diff
@@ -0,0 +1,40 @@
+import { registerDecorator, ValidationOptions } from 'class-validator';
+import { WebhookEventType } from './webhook-event-types';
+
+export function NoDuplicateAndOtherTypes(
```
rad20c, line 4: If you follow [this comment](https://github.com/KubaBaniak/RecipeAppNestjs/pull/64/files#r1360468713), then you will end up removing this decorator.

src/webhook/webhook.repository.ts
```diff
@@ -13,14 +13,15 @@ export class WebhookRepository {
...
-    return this.prisma.webhook.create({
+    const test = await this.prisma.webhook.create({
       data: {
         ...webhookData,
         userId,
         initVector: iv,
         authTag,
       },
     });
+    return test;
```
rad20c, line 24: this is probably left over from some internal testing right? 😉

rad20c (changes requested):
see comments

src/webhook/dto/create-webhook-request.ts
```diff
@@ -1,23 +1,33 @@
...
+    message: 'name should be between 4 and 255 characters long',
+  })
   @ApiProperty()
   name: string;
 
-  @IsNotEmpty()
-  @IsString()
+  @IsArray()
+  @IsString({ each: true })
+  @Transform(({ value }) => Array.from(new Set(value)))
```
rad20c, line 23: good use of `Set` 👍🏻

rad20c (approved):

PR #66, opened 2023-10-19
docs: update webhook creation docs

KubaBaniak:


src/webhook/dto/create-webhook-request.ts
```diff
@@ -23,7 +23,9 @@ export class CreateWebhookRequest {
   @Transform(({ value }) => Array.from(new Set(value)))
   @IsIn(['recipe_created', 'recipe_updated', 'recipe_deleted'], { each: true })
   @ApiProperty({
-    enum: ['recipe_created', 'recipe_updated', 'recipe_deleted'],
+    type: String,
+    isArray: true,
+    example: ['recipe_created', 'recipe_updated', 'recipe_deleted'],
```
rad20c, line 28: how about using something like this (warning, might not be a valid code):

```typescript
type: 'array',
     items: {
            type: 'string'
         }
```
?

rad20c (approved):

src/webhook/dto/create-webhook-request.ts
KubaBaniak: Yep, it works as expected

PR #69, opened 2023-10-24
fix: reset password form

KubaBaniak:
This form should be on frontend-side and should send request to my backend, but because I don't have one, i had to hard code url

### DIFF
--- a/src/utils/reset-password-form.html
+++ b/src/utils/reset-password-form.html
@@ -87,6 +87,7 @@
     </div>
 
     <script>
+      console.log(`hello app ${process.env.APP_PORT}`);
       document
         .getElementById('resetPasswordForm')
         .addEventListener('submit', function (e) {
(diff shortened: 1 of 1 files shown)


src/utils/reset-password-form.html
```diff
@@ -112,7 +112,7 @@ <h2>Password Reset</h2>
             newPassword: newPassword,
           };
 
-          fetch('http://localhost:3000/auth/reset-password', {
+          fetch('http://13.49.21.92:3000/auth/reset-password', {
```
rad20c, line 115: this shouldn't be hardcoded - use environment variables

rad20c (changes requested):
check comment

src/utils/reset-password-form.html
rad20c: I mean the server IP / domain address of course (+ port)

rad20c: you might also want to include `http://` too, as on the server you should use `https://`, while locally localhost most probably runs on `http://`

KubaBaniak: As I mentioned in description, this is 'raw' html. And i think i cannot pass env variables to it (this form is being hosted using Apache HTTP Server)

PR #70, opened 2023-10-30
refactor: move auth servcie to other microservice

KubaBaniak:
## PR with microservice changes: https://github.com/KubaBaniak/RecipeAppAuth/pull/14

Im going to take care of tests in next PR (https://github.com/KubaBaniak/RecipeAppNestjs/pull/71)


src/auth/auth.module.ts
```diff
@@ -25,22 +23,29 @@ import { PendingUsersRepository } from '../user/pending-user.repository';
...
     }),
     MailModule,
+    RabbitMQModule.forRoot(RabbitMQModule, {
+      exchanges: [
+        {
+          name: 'authentication',
+          type: 'topic',
+        },
+      ],
+      uri: 'amqp://127.0.0.1:5672',
```
rad20c, line 33: this should perhaps be in env variable

src/auth/auth.service.ts
```diff
@@ -1,59 +1,35 @@
...
-  }
-
-  async generateToken(
-    id: number,
-    secret: string,
-    time?: string,
-  ): Promise<string> {
-    const payload = { id };
-    const options: JwtSignOptions = { secret };
+  private EXCHANGE = 'authentication';
```
rad20c, line 22: hmm, I think this should not be in capital letters, just simply `exchange`

src/auth/auth.service.ts
```diff
@@ -1,59 +1,35 @@
...
 
-    if (time) {
-      options.expiresIn = time;
+  validateAuthMicroserviceReturn(payload: any) {
```
rad20c, line 24: `any` warning - if payload can really be anything, it's better to use `unknown`. However I see you are expecting `message` and `status` fields inside, so there must be some certain type passed here. If there can be multiple types passed here then you can use unions, e.g. `TypeA | TypeB`. If inside the type there are few fields defined and there can be other unknown fields, you can use something like this: `[key: string]: unknown` (example [here](https://johndpelingo.com/notes/typescript-known-and-unknown-keys/))

src/auth/auth.service.ts
```diff
@@ -1,59 +1,35 @@
...
+    if (typeof payload === 'object' && 'status' in payload) {
```
rad20c, line 25: This looks like you are checking if the response is of some ErrorType. If you define the types you are expecting here, you can use a "type guard" to check if it's error or not. See [this article](https://blog.logrocket.com/how-to-use-type-guards-typescript/#custom-typeguard-with-predicate) (Custom type guard with predicate section)

src/auth/auth.service.ts
```diff
@@ -77,71 +72,81 @@ export class AuthService {
...
-  }
+  async activateAccount(token: string): Promise<UserPayloadRequest> {
+    const payload = { token };
+
+    const userId = await this.amqpConnection.request<number>({
+      exchange: this.EXCHANGE,
+      routingKey: 'activate-account',
+      payload,
+    });
+    this.validateAuthMicroserviceReturn(userId);
```
rad20c, line 148: I like to put 1 empty line in such cases (after constructs that take more than 1 line)

rad20c (approved):

PR #71, opened 2023-11-03
refactor: tests with microservice

KubaBaniak:
## Refactor tests for this PR: https://github.com/KubaBaniak/RecipeAppNestjs/pull/70


.circleci/config.yml
```diff
@@ -8,6 +8,11 @@ defaults: &defaults
     - image: cimg/node:18.16.0
     - image: cimg/postgres:14.1
     - image: redis:7.0.12-alpine
+    - image: rabbitmq:3.13.0-rc.1-alpine
```
rad20c, line 11: I wouldn't use release candidate version of images - I'd recommend switching to `3.12.8-alpine`

src/user/test/user.service.spec.ts
```diff
@@ -28,15 +28,16 @@ describe('UserService', () => {
   describe('CreateUser', () => {
     it('should create default User', async () => {
       //given
-      const request = createUser();
+      const { email } = createUser();
+      const password = faker.internet.password({ length: 64 });
 
       //when
-      const createdUser = await userService.createUser(request);
+      const createdUser = await userService.createUser({ email, password });
```
rad20c, line 35: hmm, why did we remove the `password` from `createUser` if we need it anyway?

src/auth/test/auth.service.spec.ts
```diff
@@ -201,106 +131,110 @@ describe('AuthService', () => {
...
+      const generatedRecoveryKeys = await authService.generate2faRecoveryKeys(
+        userId,
+      );
+
+      expect(generatedRecoveryKeys).toEqual(recoveryKeys);
     });
   });
 
   describe('Enable 2fa', () => {
     it('should enable 2fa', async () => {
```
rad20c, line 202: You could do `ctrl + f` in IDE and find all instances of `"2fa"` as strings / descriptions etc. and change them to `"2FA"` ("2FA" in strings, "2fa" in object / class / variable names etc.)

test/webhook.e2e-spec.ts
```diff
@@ -8,23 +8,23 @@ import { User } from '@prisma/client';
...
 describe('WebhookController (e2e)', () => {
   let app: INestApplication;
-  let authService: AuthService;
   let prismaService: PrismaService;
   let user: User;
   let accessToken: string;
+  let jwtService: JwtService;
 
-  beforeAll(async () => {
+  beforeEach(async () => {
```
rad20c, line 27: Why did you change it to `beforeEach`? I think testing module should be created once. Now you are creating it for each test.

rad20c (approved):

src/user/test/user.service.spec.ts
KubaBaniak: I used this function a lot while creating data for auth tests. But now since the `db` here (of base service) does not contain user password, and most of the endpoints did not required password after that change I removed it.

test/webhook.e2e-spec.ts
KubaBaniak: I've noticed that I did not have a call to database that deletes all webhooks and its events before. So to fix that, what I did is change it so it runs multiple testing modules. 

But I changed it, so test module is created once, and records created in test are being deleted after.

